PR #19, opened 2021-06-28
document id parameter

nathanstitt:
Otherwise swagger complains:
```
Errors:
	-attribute paths.'/research/studies/{id}/stages'. Declared path
  parameter id needs to be defined as a path parameter in path or
  operation level
  ```

### DIFF
--- a/backend/app/controllers/api/v0/researcher/stages_swagger.rb
+++ b/backend/app/controllers/api/v0/researcher/stages_swagger.rb
@@ -77,6 +77,13 @@ class Api::V0::Researcher::StagesSwagger
         'Studies'
       ]
       parameter do
+        key :name, :id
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: backend/app/controllers/api/v0/researcher/stages_swagger.rb [hunk 1]
```diff
@@ -76,6 +76,13 @@ class Api::V0::Researcher::StagesSwagger
       key :tags, [
         'Studies'
       ]
+      parameter do
+        key :name, :id
+        key :in, :path
+        key :description, 'The stage ID'
```
jpslav, line 82: Should be study ID

jpslav (changes requested):

jpslav (approved):

PR #23, opened 2021-07-07
UI to add a study

nathanstitt:
image


backend/app/controllers/application_controller.rb
```diff
@@ -17,7 +19,7 @@ def validate_not_real_production
 
   def current_user_uuid
     @current_user_uuid ||=
-      if Rails.env.development?
+      if Labs.use_cookie_authentication?
```
jpslav, line 22: All authentication is cookie authentication (the `else` below is SSO cookie authentication) so this may not be the right name.  I get that below we're reading the header manually so not really using the "cookie" features.  

`Labs.allow_stubbed_authentication?`?

jpslav (changes requested):

backend/app/controllers/application_controller.rb
nathanstitt: yes that makes sense, I'll update

nathanstitt: hm, given that all auth is cookie authentication, then we probably always want to use cookies, and we'll only need to test `dev || test` in one spot, which in turn removes the desire to DRY up the logic. I'll just check that in current_user_uuid

jpslav (approved):

PR #33, opened 2021-08-18
Add Heroku deployment

nathanstitt:
It's running at https://labs-web.herokuapp.com/ but is not able to read the OX cookie due to not being hosted under the OpenStax.org domain.  More investigation is needed to find the right approach.


reedstrm (changes requested):
looks reasonable to me. Let's hold on merging until we  (you) solve thre cookie issue, so this PR will then complete the  feature "Heroko dev/testing deployment"

PR #34, opened 2021-08-18
upgrade Ruby to 2.6.8 - various package upgrades

reedstrm:
What that says up there. 2.6.8 is supported by Github Workflows BTW.


nathanstitt:
how about we try to move to rails 3.x and see how that works?

PR #38, opened 2021-08-31
update to ruby 3 and fix specs/docker build

nathanstitt:


backend/spec/support/api_v0_helpers.rb
```diff
@@ -72,19 +72,23 @@ def clear_headers
   end
 
   def api_post(*args, &block)
-    post(*prep_request_args(args), &block)
+    url, opts = prep_request_args(args)
+    post(url, **opts, &block)
```
reedstrm, line 76: THis is to allow double-de-nesting of the return from prep_request_args, I take it? (If there's more than one opt, basically)

reedstrm (approved):

backend/spec/support/api_v0_helpers.rb
nathanstitt: yes, it's because Ruby 3 doesn't allow expanding both keyword and positional arguments using *,.  get/post/put/delete etc have the first as positional  and the 2nd argument is keyword

https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

PR #40, opened 2021-09-01
Implement Heroku review apps w/ custom openstax domain

reedstrm:
Using a new robot user (heroku-labs) that can only assume one role (research-labs-dns) that can only do a limited sot of operations on the labs[.sandbox].openstax.org hosted zones. This allows automated creation of deployments for every PR, under an openstax.org domain, so all the cookie work!

### DIFF
--- a/Dockerfile.heroku
+++ b/Dockerfile.heroku
@@ -34,6 +34,8 @@ ENV RUBY_ENV=production
 ENV RUBY_SERVE_STATIC_FILES=true
 COPY --from=node /code/frontend/dist/ ./public/
 
+# Extra comment, just to create a PR
+
 ARG SENTRY_RUBY_DSN
 ENV RAILS_ENV=production
 CMD bundle exec rake about && bundle exec rails server -b '0.0.0.0' -p $PORT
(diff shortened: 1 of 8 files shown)


reedstrm:
sigh need to bounce PR to test scripts.

reedstrm:
bouncy bouncy

reedstrm:
yup, again

Dockerfile.heroku
```diff
@@ -34,6 +34,8 @@ ENV RUBY_ENV=production
 ENV RUBY_SERVE_STATIC_FILES=true
 COPY --from=node /code/frontend/dist/ ./public/
 
+# Extra comment, just to create a PR
```
Dantemss, line 37: Delete? :)

backend/lib/tasks/review_app_setup.rake
```diff
@@ -0,0 +1,74 @@
...
+          ],
+          comment: "Review domain for labs #{subdomain}"
+        }
+      })
+
+    change_id = change.change_info.id
+
+    # Wait for change to be active
+    delay = 1
+    while (change.change_info.status == 'PENDING') & (delay < 30)
```
Dantemss, line 64: Not sure if the status is guaranteed to be pending if you check immediately? Hopefully...

Dantemss: Also `&` with booleans is non-short-circuit which I never see used but I guess it works.

Dantemss (approved):

reedstrm:
FE test fails on timeout, known issue will fix separately

PR #41, opened 2021-09-01
support launching as a preview

nathanstitt:
Previews do not record launch or viewed state on qualtrics

image


backend/app/services/launch_pad.rb
```diff
@@ -7,26 +7,20 @@ def initialize(study_id:, user_id:)
     @user_id = user_id
   end
 
-  def launch
+  def launch_url(preview: false)
     raise(LaunchError, 'This study is not open.') unless study.open?
```
reedstrm, line 11: Shouldn't we allow researchers to preview studies that are not yet open?

backend/app/services/launch_pad.rb
```diff
@@ -7,26 +7,17 @@ def initialize(study_id:, user_id:)
...
-        else
-          raise "Unsupported stage type: '#{launch_config[:type]}'"
-        end
-
-      launcher.new(config: launch_config, user_id: user_id).url
-    end
-
-    url || raise('An error occurred when building a launch url')
+    if preview
+      study.stages.order(:order).first.launcher(user_id).url + '&Q_CHL=preview&Q_SurveyVersionID=current'
```
reedstrm: Hmm, this is a Qualtrix-specific method of making a preview URL - knowledge of qualtrix-ness is buried in the launcher, so I think we should extend that model to have a preview_url component, instead?

reedstrm (changes requested):
Huh, thought this preview work was going to be a frontend thing. From what I can see, all the frontend work is changing implementation of some bit from React to JSX, and lint work?

Two Qs.

backend/app/services/launch_pad.rb
nathanstitt: yes, I suppose we might as well allow that. shouldn't hurt anything and I can see where it'd be useful

backend/app/services/launch_pad.rb
nathanstitt: good catch, yep this should live in qualtrics launcher.  other launchers might not support preview and could raise an exception or something

reedstrm (approved):

PR #42, opened 2021-09-03
Open previews in modal and close it when study completed

nathanstitt:
The landing page needed to communicate through the iframe back to it's parent and close the modal


reedstrm (approved):
Interesting to see the model evolve from MandatoryModal to StudyModal.

PR #53, opened 2021-09-09
Disable launching and details on completed studies

nathanstitt:
Once a study's completed it should appear disabled and not accept click events


reedstrm (approved):
Just for my edification, what's the diff between container and container-fluid classes?

nathanstitt:
`container-fluid` is full width, it was causing the buttons on the navbar to expand wider than the page content and look a little strange

nathanstitt:
FYI here's the docs: https://getbootstrap.com/docs/5.1/layout/containers/#how-they-work

PR #54, opened 2021-09-10
allow deletion of  studies if they've not been used

nathanstitt:
Will allow researchers to remove a study if they make a mistake setting it up or something


reedstrm (approved):
Yay! Used the review app to verify "yup, it worked!" However, we've still got issues w/ the frontend spec/tests. Can you take one more look at that adn see if anything jumps out?

PR #57, opened 2021-09-16
Add ResearchId model

nathanstitt:
Adds model to store a unique random id for research participants

That id is what we will send to outside agencies so that if it's compromised, it cannot be linked back to a real OX user


reedstrm (approved):
Looks good!

PR #63, opened 2021-09-28
add demo-users rake command

nathanstitt:
Sets up a few demo accounts for use in automated testing


reedstrm (approved):
Works a treat! Added to review app workflow, so the users always exist

PR #64, opened 2021-09-29
Kenetic rename and domain setting

reedstrm:
Fixed up names in the review app post-deploy rake scripts
Added a rake script to set domains for production heroku deployments, as well.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,9 +24,9 @@ jobs:
         run: |
           sleep 30
           docker ps -a
-          docker logs labs_api_1
+          docker logs kenetic_api_1
           curl -i http://localhost:4006/api/v0/swagger.json
-          docker logs labs_api_1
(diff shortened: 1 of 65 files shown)


reedstrm:
Looks like there's some docker "labs => kenetic" fixes needed. On it!

nathanstitt (commented):
what happens if this is re-ran after the domains are already setup?  Does it error or anything?

reedstrm:
Should be idempotent - 

> what happens if this is re-ran after the domains are already setup? Does it error or anything?

Yeah, should probably make that idem-potent. The AWS part will be fine, since it's an UPSERT/changset. If the value of the CNAME is the same, it'll still do it. The heroku parts not so much, I'll take a crack at wrapping some test-and-set logic around those (Just tested `heroku domains:add -a kenetic-web staging.kenetic.openstax.org` throws an error.)

nathanstitt (approved):

PR #79, opened 2021-10-19
Tagging

nathanstitt:
Still to be done is some sort of filtering on the participant side, and they'll be additional tags pre-loaded as options in the models/study.ts file

image

image


reedstrm (approved):
Looks good for what it does. Concern, are we constraining that there must be a tag named category, with one of the three allowed types somewhere, or did we decide that was too restrictive? I see we require _some_ sort of tag, but not constraining it's type or values.

nathanstitt:
Yes, it's currently requiring a `type` tag (models/study.ts) but we'll probably extend that validation further to include subjects once that's added

PR #81, opened 2021-11-04
Add qualtrics api and use survey_id 

nathanstitt:
We need the survey id to retrieve data from qualtrics.  Since it's only unchanging part of the url we can build the survey from the id


backend/app/services/qualtrics_launcher.rb
```diff
@@ -3,15 +3,17 @@
...
+  def initialize(secret_key:, survey_id:, user_id:, study_id:)
+    @secret_key = secret_key
+    @survey_id = survey_id
     @user_id = user_id
     @study_id = study_id
   end
 
   def url
-    uri = URI(config[:url])
+    uri = URI("#{Rails.application.secrets.qualtrics_launch_url}/#{survey_id}")
```
reedstrm, line 16: This does lock us in on using our qualtrics (or at least, only 1 qualtrics account) for the entire app. The previous "give us the whole url" did allow for using arbitrary qualtrics accounts. Fine for now, but perhaps we should use the launch url from secrets/config as the default for an optional initialization param, so some surveys can have different accounts in the future.

reedstrm (changes requested):

backend/app/services/qualtrics_launcher.rb
nathanstitt: good catch 😁 

Yes, I thought of that, but decided against it for right now.  in the future when we use more of the qualtrics api we'll probably be using oauth with it, because that will allow a teacher to grant us access to their account by redirecting to qualtrics and granting us access.

When that happens we'll need a "qualtrics" profile associated with researchers and will no longer use the account specific urls at all.

reedstrm: 👍🏻

nathanstitt:
@reedstrm can you approve when you're able? 🥺

reedstrm (approved):

PR #82, opened 2021-11-05
Rearrange files and split views

nathanstitt:
should make things a little more efficient loading


reedstrm (approved):
Looks to be mostly adding the menu, and some refactor.

PR #83, opened 2021-11-08
Fix sentry logging

nathanstitt:
needs dsn while building and also `production` environment should be symbol


reedstrm (approved):
Looks fine. I used rubocop to fix the minor lint problem that was blocking.

PR #86, opened 2021-11-10
Participant filtering and sort

nathanstitt:
Mobile:
image

Desktop:
image


frontend/src/screens/studies/participant-controls.tsx
```diff
@@ -0,0 +1,195 @@
...
+            input: {
+                marginRight: '0.5rem',
+            },
+        }}>
+            {Object.entries(StudySubjects).map(([value, label]) => (
+                <label key={value} className="dropdown-item">
+                    <input
+                        type="checkbox"
+                        value={value || ''}
+                        checked={state.subjects?.includes(value as any) || false}
```
RoyEJohnson, line 72: Isn't "|| false" redundant?

frontend/src/screens/studies/participant-controls.tsx
```diff
@@ -0,0 +1,195 @@
...
+        } else if (sort == 'time:low-high') {
+            studies = sortBy(studies, 'durationMinutes')
+        } else if (sort == 'points:high-low') {
+            studies = sortBy(studies, s => (s?.participationPoints || 0)* -1)
+        } else if (sort == 'points:low-high') {
+            studies = sortBy(studies, 'participationPoints')
+        }
+    }
+    if (subjects?.length) {
+        studies = studies.filter(s => subjects.find(subj => -1 != s.tags.indexOf(subj)))
```
RoyEJohnson, line 128: I'd write `find(subj => s.tags.includes(subj)` here, I think.

RoyEJohnson (approved):
Interesting to review! You get a lot of mileage out of Box; I will have to look at that. Also will look at Iconify and Rooks. Thanks for inviting me to review.

frontend/src/screens/studies/participant-controls.tsx
nathanstitt: hehe, yeah it is, but if it returns undefined and then later false/true, react complains about the input changing from `uncontrolled` to `controlled`.  It's better to consistently return the same type

frontend/src/screens/studies/participant-controls.tsx
nathanstitt: good point, didn't think of that and reached for the old `indexOf` instead. I'll update, thank you!

nathanstitt:
@RoyEJohnson thanks for the review.   yes I like the simplicity of how using Box with the props lets you roughly grasp how things are laid out.  Not sure it'll work for you as-is though since it uses emotionjs internally to apply the styles.

PR #87, opened 2021-11-11
set env vars and link to homepage

nathanstitt:
needs to read the homepage_url from the backend config so it's set at run time. heroku doesn't expose the config vars when the docker image is built otherwise we could encode it there


frontend/src/api/models/Environment.ts
```diff
@@ -25,6 +25,12 @@ export interface Environment {
      * @memberof Environment
      */
     readonly accountsEnvName?: string;
+    /**
```
reedstrm, line 28: ??

reedstrm (approved):

PR #88, opened 2021-11-12
extract tags into own file; add a few new ones

nathanstitt:


reedstrm (approved):
simple refactor, looks good.

PR #89, opened 2021-11-19
add homepage to repo so it can be tracked

nathanstitt:
It's plain html that gets copy-n-pasted into wagtail, but is easier to develop locally then use the `copy-to-clipboard` script to grab the contents


nathanstitt:
@RoyEJohnson I listed you as reviewer just in case you were interested in how I build it, nothing really to review though since it's already live, this is just a place for me to store the file so it's easier to modify later

PR #90, opened 2021-11-22
set modal and iframe to be full screen

nathanstitt:
Screen Shot 2021-11-22 at 11 16 03 AM


reedstrm (approved):
looks fine, though was this overridden by later decisions/requests from Chris? ISTR something about 80% except on tablet and below, not seeing that here.

PR #92, opened 2021-12-01
Api method to determine eligibility

nathanstitt:


nathanstitt:
@TomWoodward heads up this is the PR to add the eligibility endpoint that will live at `/kinetic/api/v1/eligibility` 

It accepts a URL encoded cnx book slug in the `book` query parameter, and will also read the cloudfront geo location header

reedstrm (approved):
Seems reasonable. Might need to get more complicated in the future, but that's for the future.

PR #98, opened 2021-12-09
move CORS config to initializer

nathanstitt:


backend/config/initializers/cors.rb
```diff
@@ -1,17 +1,31 @@
...
-#
-#     resource '*',
-#       headers: :any,
-#       methods: [:get, :post, :put, :patch, :delete, :options, :head]
-#   end
-# end
+Rails.application.config.middleware.insert_before 0, Rack::Cors do
+  allow do
+    origins '*'
+    resource '/api/v0/swagger*', headers: :any, methods: [:get, :options, :head]
```
reedstrm, line 13: shouldn't this be v1?

reedstrm (changes requested):
api version update

backend/config/initializers/cors.rb
nathanstitt: yep it should be, actually I'm not sure why we have CORS on for swagger docs.  I don't think it's needed

PR #99, opened 2021-12-09
Generate a simple CSV activity report

nathanstitt:
also adds a accounts key/secret so it can read the usernames/emails


reedstrm (approved):
Looks good for as far as it goes, which is as far as we can, for now. We can add the teacher details once we actually collect them from students. :)

PR #100, opened 2021-12-10
log exhaustive details about request

nathanstitt:


reedstrm:
This was useful, but it's not intended for production, correct?

nathanstitt:
yes, we can merge it but need to take out the `around` line in the eligibility controller first

reedstrm (approved):
more logging available is good :)

PR #105, opened 2022-01-07
Track and display progress toward rewards

nathanstitt:
image


backend/app/bindings/api/v1/bindings/environment.rb
```diff
@@ -49,18 +54,39 @@ def initialize(attributes = {})
...
+
+      if @homepage_url.nil?
+        invalid_properties.push('invalid value for "homepage_url", homepage_url cannot be nil.')
+      end
+
+      if @rewards_schedule.nil?
+        invalid_properties.push('invalid value for "rewards_schedule", rewards_schedule cannot be nil.')
+      end
+
       invalid_properties
```
reedstrm, line 81: V.nice. I love apis that explain "Invalid!!!"

backend/config/rewards_schedule.yml
```diff
@@ -0,0 +1,20 @@
...
+-
+  prize: $50 Nike gift card
+  points: 50
+  start_at: 2022-03-15
+  end_at: 2022-04-14
+-
+  prize: Apple AirPods™
+  points: 0
+  start_at: 2022-04-20
+  end_at: 2022-04-20
```
reedstrm, line 20: This is read on startup, so can (and should) be overridden by the deployment, right?

reedstrm (approved):

backend/config/rewards_schedule.yml
nathanstitt: no, this is the "real" schedule.  eventually we'll store it in the DB and write an admin UI to allow modifications

frontend/src/lib/api-config.ts
```diff
@@ -2,15 +2,15 @@ import { useMemo } from 'react'
 import { ENV } from './env'
 import { Configuration, StudiesApi } from '../api'
 
-export const useStudyApi = () => {
 
+export const ApiConfiguration = new Configuration({
```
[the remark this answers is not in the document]
nathanstitt: Nope, good catch, that should probably be a lowercase leading char, `apiConfiguration` or probably better as all caps to denote a constant: `API_CONFIGURATION`

frontend/src/models/rewards.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export interface CalculatedRewardsScheduleSegment extends RewardsScheduleSegment {
+    totalPoints: number
+    index: number
+}
+
+export function rewardPointsEarned(studies: ParticipantStudy[]): number {
+    return studies.reduce((points, study) => {
+        if (study.participationPoints && study.completedAt) {
+            points += study.participationPoints
```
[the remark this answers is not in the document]
nathanstitt: yes, mutating doesn't make sense since it's using reduce, it just needs to return the value.  I do like an if vs the ternary though since we shouldn't do anything if it's not completed.

PR #106, opened 2022-01-13
record metadata returned from study landings

nathanstitt:
Adds a new table to contain arbitrary json from surveys


nathanstitt:
@RoyEJohnson this includes the fixes you suggested from #105 - thanks!

reedstrm (approved):

PR #108, opened 2022-01-14
allow origins that REX uses to load eligibility api

nathanstitt:


reedstrm (commented):
The logic here is not obvious, so at some point we may want to put in some comment/notes about _why_ we're allowing localhost urls, and maybe not do it in production?

PR #109, opened 2022-01-14
use regex for rack:ors

nathanstitt:
The earlier #108 didn't work, and reading https://github.com/cyu/rack-cors carefully there's no mention of using wildcards with strings on origin checks, it suggests regex


reedstrm (approved):
Let's try it

PR #112, opened 2022-01-18
Fix non consent

nathanstitt:
The param name was incorrect and the landing class had some logic errors

### DIFF
--- a/backend/app/services/launch_pad.rb
+++ b/backend/app/services/launch_pad.rb
@@ -37,9 +37,8 @@ class LaunchPad
   end
 
   def abort(reason)
-    stage = stage_to_land
     if reason == 'refusedconsent'
-      stage.refuse_to_consent!
+      launched_study.refuse_to_consent!
       return true
     end
 
--- a/frontend/src/screens/study-landing.tsx
+++ b/frontend/src/screens/study-landing.tsx
@@ -25,7 +25,7 @@ export default function UsersStudies() {
     const api = useStudyApi()
     const history = useHistory()
     const user = useCurrentUser()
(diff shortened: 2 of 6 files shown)


frontend/src/screens/study-landing.tsx
```diff
@@ -25,7 +25,7 @@ export default function UsersStudies() {
     const api = useStudyApi()
     const history = useHistory()
     const user = useCurrentUser()
-    const noConsent = useQueryParam('consent') == 'true'
```
Dantemss, line 28: I'd say throw an error if it is neither true nor false. Maybe the error can cause a 4xx reply though.

reedstrm (approved):

backend/app/services/launch_pad.rb
```diff
@@ -37,9 +37,8 @@ def land
   end
 
   def abort(reason)
-    stage = stage_to_land
     if reason == 'refusedconsent'
```
Dantemss, line 40: Same thing here, instead of silently continuing if you receive an unexpected reason, why not throw an error?

Dantemss (commented):
Just a couple of comments

frontend/src/screens/study-landing.tsx
nathanstitt: good point but usually it's omitted altogether.  the param's only set if the user left the survey early

backend/app/services/launch_pad.rb
nathanstitt: yes that's a good idea. I think I'll update to do so.  Thanks!

PR #114, opened 2022-01-19
Add analytics lib, track using both GA and tag manager

nathanstitt:
Also support test and prod ga configs


jivey (approved):
Looks good to me 👍

PR #119, opened 2022-01-27
prevent duplicate launches

nathanstitt:
If the button was double clicked the launches could occur concurrently and cause a:
> PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_launched_studies_on_user_id_and_study_id"
DETAIL:  Key (user_id, study_id)=(a05196be-0400-4ff3-a0ec-912ce1f568d5, 2) already exists.

https://sentry.io/organizations/openstax/issues/2967634068/?environment=production&project=6051748

### DIFF
--- a/frontend/src/screens/study-details.tsx
+++ b/frontend/src/screens/study-details.tsx
@@ -10,13 +10,25 @@ import dayjs from 'dayjs'
 
 const LaunchStudyButton: React.FC<{ study: ParticipantStudy }> = ({ study }) => {
     const api = useStudyApi()
+    const [isBusy, setBusy] = useState(false)
+    const onLaunch = async () => {
+        setBusy(true)
+        await LaunchStudy(api, study)
+        setBusy(false)
+    }
     if (study.completedAt) {
         return (
             <b>Completed on {dayjs(study.completedAt).format('LL')}</b>
         )
     }
     return (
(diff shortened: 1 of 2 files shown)


frontend/src/screens/study-details.tsx
```diff
@@ -10,13 +10,25 @@ import dayjs from 'dayjs'
 
 const LaunchStudyButton: React.FC<{ study: ParticipantStudy }> = ({ study }) => {
     const api = useStudyApi()
+    const [isBusy, setBusy] = useState(false)
+    const onLaunch = async () => {
+        setBusy(true)
```
Dantemss, line 15: As far as I can tell this is asynchronous so if the UI is slow enough you can probably still double-click before it disables?

Dantemss (commented):
Solving it in the UI may be good enough if you can figure it out. But if not, you can probably solve it in the backend with something like: https://stackoverflow.com/a/60443582

It's probably the only way to solve it when there are multiple people using the same account or other shenanigans.

frontend/src/screens/study-details.tsx
nathanstitt: well, it's complicated :)   setState is indeed async, BUT react will call it first before re-rendering and accepting any input events.  That has the effect of making sure the button is disabled via the 'busy' prop before any new browser click events

Dantemss (approved):

PR #126, opened 2022-02-14
Limit sampling to the api eligibility endpoint

nathanstitt:
It's called from REX which has huge traffic and exhausted our monthly transactions limit.


reedstrm (approved):
There's a sentry helper in frontend, as well, do we need to set the sample rate there. as well? (I realize it's not getting hit by rex, just for consistency)

PR #128, opened 2022-02-15
Fix rake task and rename open to available?

nathanstitt:
open is also a scope and was triggering a warning


reedstrm (approved):
naming things is hard!

PR #129, opened 2022-02-15
add consent param, move metadata -> md

reedstrm:


frontend/specs/participant.spec.ts
```diff
@@ -95,7 +95,7 @@ test('launching study and aborting it', async ({ page }) => {
     await page.click('testId=launch-study')
 
     // qualtrics will redirect here once complete
-    await goToPage({ page, path: `/study/land/${studyId}?consent=false`, loginAs: 'user' })
+    await goToPage({ page, path: `/study/land/${studyId}?abort=true`, loginAs: 'user' })
```
nathanstitt, line 99: see comments below in study-landing.tsx, I don't think we want to rename this

nathanstitt (changes requested):

frontend/src/screens/study-landing.tsx
```diff
@@ -74,8 +74,9 @@ export default function UsersStudies() {
     const api = useStudyApi()
     const history = useHistory()
     const user = useCurrentUser()
+    const abort = useQueryParam('abort') == 'true'
     const noConsent = useQueryParam('consent') == 'false'
```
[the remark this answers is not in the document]
reedstrm: No, consent is still the primary, but it doesn't not control if the study was aborted - that's now explicit

frontend/specs/participant.spec.ts
reedstrm: Lack of consent is no longer an abort, so we need a separate way to report an aborted study. My one thought was it might be good for it to report a reason, rather than just a boolean, but we can no longer abort on consent=false

frontend/src/screens/study-landing.tsx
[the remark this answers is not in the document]
reedstrm: Line 78 - noConsent should be false unless consent exists, and is false, right? And I invert it (!noConsent) when setting the param (line 98 or so) so consent will default to true.

backend/app/controllers/api/v1/participant/studies_controller.rb
```diff
@@ -37,19 +37,19 @@ def launch
...
-    if params[:metadata]
+    if params[:md]
       ParticipantMetadatum.create!(
         user_id: current_user.id,
         study_id: @study.id,
-        metadata: params[:metadata]
+        metadata: params[:md]
       )
     end
     if params[:aborted]
```
nathanstitt, line 47: I think we'll need to check if `params[:aborted] == 'true'` here instead of just if it's present, because it could be 'false'

frontend/src/screens/study-landing.tsx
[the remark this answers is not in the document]
nathanstitt: yep, you're correct. I'd missed that noConsent was still being passed to the backed below. Sorry

nathanstitt (approved):

backend/app/controllers/api/v1/participant/studies_controller.rb
nathanstitt: yeah that's fair, but we should probably check that it's not `false` at least, that way we don't mark it as aborted when `abort=false`

PR #132, opened 2022-02-18
default abort flag to inverse of consent

nathanstitt:
This allows qualtrics study to continue with the behavior of passing consent=false and studies not being marked as completed


reedstrm (approved):
Not sure this will survive the long-term fleshing out of multi-step and consent, but I agree that abort needs to be fleshed out through the stack, anyway, at this is a reasonable default for the next bit.

PR #135, opened 2022-02-28
retake  - for opt-in, primarily

reedstrm:
allow re-launching if opted out

### DIFF
--- a/backend/app/models/launched_study.rb
+++ b/backend/app/models/launched_study.rb
@@ -12,11 +12,13 @@ class LaunchedStudy < ApplicationRecord
   end
 
   def retakeable?
-    completed_at == opted_out_at
+    # Timestamps are in microseconds - consider 0.1 sec window as the "same"
+    completed_at - opted_out_at < 0.1
   end
 
   def completed!(consent: true)
-    update!(completed_at: Time.now, consent_granted: consent)
+    update!(completed_at: Time.now, consent_granted: consent,
+            opted_out_at: consent ? nil : Time.now)
   end
 
   def aborted!
--- a/backend/app/models/launched_study.rb
+++ b/backend/app/models/launched_study.rb
@@ -14,7 +14,7 @@ class LaunchedStudy < ApplicationRecord
   def retakeable?
     completed_at == opted_out_at
   end
-  
+
   def completed!(consent: true)
     update!(completed_at: Time.now, consent_granted: consent)
   end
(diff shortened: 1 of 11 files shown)


frontend/src/models/study.ts
```diff
@@ -56,7 +56,7 @@ export const LaunchStudy = async (api: StudiesApi, study: {id: number}, options:
 
 export const isStudyLaunchable = (study: ParticipantStudy) => {
     return Boolean(
-        !study.completedAt &&
+        (!study.completedAt || (dayjs(study.completedAt).isSame(dayjs(study.optedOutAt), 'second'))) &&
```
reedstrm, line 59: Had trouble comparing these timestamps in ruby, they differed in the millisecond range, so I went with a delta comparison method. That is more robust to "splitting the second", think I should use it here?

reedstrm (commented):
Smoe design-decision notes. :)

backend/app/models/user.rb
```diff
@@ -34,9 +34,15 @@ def launch_next_stage!(study)
     else
       completed_stage_ids = launched_stages(study: study).map(&:stage_id)
       stage = study.stages.where.not(id: completed_stage_ids).order(:order).first
-      raise 'No stage to launch exists' if stage.nil?
-
-      launch = LaunchedStage.create(stage_id: stage.id, user_id: id)
+      if stage.nil?
+        # See if study is retakable
+        # Don't expect more than one, just unwrapping the returned set
+        launched_study = LaunchedStudy.where({ user_id: id, study_id: study.id }).first
```
reedstrm, line 40: going to need to rework this logic when we enable multi-step

backend/app/models/launched_study.rb
```diff
@@ -11,8 +11,14 @@ def completed?
...
   end
 
+  def retakeable?
+    # Timestamps are in microseconds - consider 0.1 sec window as the "same"
+    completed_at - opted_out_at < 0.1
+  end
+
   def completed!(consent: true)
-    update!(completed_at: Time.now, consent_granted: consent)
+    update!(completed_at: Time.now, consent_granted: consent,
```
reedstrm, line 20: I suppose the timestamps would be the same if I wrapped this in a transaction, or fetched it into a variable and used that for both - yup,  I think that's what I should do, actually.

frontend/src/models/study.ts
reedstrm: Fixed in backend by making sure they're the same  timestamp (delta code had a bug, anyway)

reedstrm:
Gong a different way.

PR #137, opened 2022-03-04
Enable multi stage studies

nathanstitt:
Adds title, description, and `available_after_days` to stages and fixes multi-stage launch logic


backend/app/models/launched_stage.rb
```diff
@@ -4,11 +4,18 @@ class LaunchedStage < ApplicationRecord
   belongs_to :stage
   belongs_to :research_id, foreign_key: :user_id, primary_key: :user_id
 
-  before_create { self.first_launched_at ||= Time.now }
+  before_validation do
+    self.first_launched_at ||= Time.now
+    self.research_id ||= ResearchId.for_user_id(user_id)
```
reedstrm, line 9: Ahhh, this is the correct fix for an issue I was seeing that I "fixed" in the specs.

reedstrm (approved):
Like the refactoring, moving logic into properly named methods.

PR #149, opened 2022-04-18
no longer auto-reload on error

nathanstitt:


nathanstitt:
closing this since it seems to be working ok, probably don't need to remove the reload

PR #152, opened 2022-05-10
Accurate completion

nathanstitt:
Return completion status when landing a stage and use it to make sure we don't say a study is completed when it is not


reedstrm (approved):
Looks good - did it exactly like I was thinking, returning status from the landing call. Was thinking it might also return which step was completed, for a multi-step, but that didn't get discussed.

PR #156, opened 2022-06-14
Only count points earned during current rewards cycle

nathanstitt:
Note this only tests against first segment, getting clarification if we should also remove "late work"


reedstrm (approved):
Seems fine. Was thrown a bit by the changes to landing, but I see it's making it a bit more robust.

PR #170, opened 2022-08-29
Save user preferences

nathanstitt:
only email now, but might be others later


Dantemss (approved):

reedstrm (commented):
Looks fine, though wondering what the use-case for is_hidden is, where the study is_open but hidden?

PR #173, opened 2022-08-30
assorted fixes for menu

nathanstitt:
Fixes a few issues that were identified:
  * no chevron on menu
  * menu should not contain link to studies
  * logout doesn't return to kinetic


nathanstitt:
this went on way too long, gonna merge it before we're tempted to add more commits

PR #175, opened 2022-09-09
Send email when various user events occur

nathanstitt:


reedstrm (approved):
I little concerned about just how thoroughly hard-coded all the email parts are, but it does seem properly factored, to allow swapping out only an class or two with different underlying implementation, if we need more flexibility in the future. Ship it. :)

PR #176, opened 2022-09-12
Adjust margin & padding surrounding multi-session

nathanstitt:
fixes https://github.com/openstax/research/issues/144

Before:
image

After:
image

### DIFF
--- a/frontend/src/components/segment-bar.tsx
+++ b/frontend/src/components/segment-bar.tsx
@@ -6,10 +6,12 @@ import { CSSObject } from '@emotion/react'
 
 interface SegmentProps {
     percentage: number
+    className?: string
 }
 
-const segmentWidth = 50
+const segmentWidth = 80
 const barWidth = 7
+const segmentTitlePadding = (segmentWidth * 0.5) - 15
 
 const segmentStyle: CSSObject = {
     position: 'absolute',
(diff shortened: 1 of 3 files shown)


frontend/src/components/segment-bar.tsx
```diff
@@ -18,18 +21,29 @@ const segmentStyle: CSSObject = {
...
+export const Segment: FCWC<SegmentProps> = ({ children, className, percentage }) => {
+    return <SegmentWrapper className={cx('segment', className)} css={{ left: `calc(${percentage}% - ${segmentWidth / 2}px)` }
+    }> {children}</SegmentWrapper >
 }
 
 export const segmentCircleStyle: CSSObject = {
-    height: '20px',
-    width: '20px',
-    borderRadius: '10px',
+    height: circleDiameter,
```
chrisbendel, line 44: Im assuming this will just interpret `20` as a value in pixels right? rather than the explicit `${circleDiameter}px`

chrisbendel (approved):

frontend/src/components/segment-bar.tsx
nathanstitt: yep, this is emotionjs, and it defaults to `px` for bare numbers.

PR #189, opened 2022-09-27
Add boxshadow back to bottom of card, fix dev logout

chrisbendel:


frontend/src/components/top-navbar.tsx
```diff
@@ -25,7 +26,7 @@ export const TopNavBar: FCWOC<{ className?: string }> = ({ children, className }
                             <Menu alignEnd >
                                 {isMobile && <li><a className="dropdown-item" href="/studies">Studies</a></li>}
                                 <li><a className="dropdown-item" href="/account">My account</a></li>
-                                <li><a className="dropdown-item" href={env.logoutURL}>Log out</a></li>
+                                <li><a className="dropdown-item" href={env.logoutURL} onClick={user.logout}>Log out</a></li>
```
nathanstitt, line 29: currently logging out in the model is a "dev env only" function.  

When deployed to staging/prod, the BE doesn't even have that controller loaded.   Also calling the `logout` function directly in the onClick will lose the user class scope and any calls to `this` inside the method will fail.  It'd need to be something like: `onClick={() => user.logout()}` or have the method bound to "this" in order to work.  I'm guessing you're not seeing that because it's loading the logoutURL and console error messages are lost.

How about changing the user method (https://github.com/openstax/kinetic/blob/main/frontend/src/models/user.ts#L59) to something like:
```ts
    logout: async (ev?: React.MouseEvent<HTMLAnchorElement>) => {
        ev?.preventDefault()
        if (ENV.IS_DEV_MODE) {
            await fetch(`${ENV.API_ADDRESS}/development/users/log_out`, {
                method: 'DELETE', credentials: 'include',
            })
            this.id = ''
            this.isAdmin = this.isResearcher = false
            window.location = '/users' // the dev user picker screen, not sure this is it
        } else {
            window.location = env.logoutURL
        }
    }
```

That probably has typo's and such, but you get the idea 😁

nathanstitt (commented):
take a look at the logout idea and see what you think

frontend/src/components/top-navbar.tsx
```diff
@@ -23,11 +20,11 @@ export const TopNavBar: FCWOC<{ className?: string }> = ({ children, className }
...
-                            {!isMobile && <a href="/studies" css={{ color: 'white', textDecoration: 'none', fontWeight: 'bold' }}>Studies</a>}
+                            {!isMobile && <Link to="/studies" css={{ color: 'white', textDecoration: 'none', fontWeight: 'bold' }}>Studies</Link>}
                             <Menu alignEnd >
-                                {isMobile && <li><a className="dropdown-item" href="/studies">Studies</a></li>}
-                                <li><a className="dropdown-item" href="/account">My account</a></li>
-                                <li><a className="dropdown-item" href={env.logoutURL} onClick={onLogout}>Log out</a></li>
+                                {isMobile && <li><Link className="dropdown-item" to="/studies">Studies</Link></li>}
+                                <li><Link className="dropdown-item" to="/account">My account</Link></li>
+                                <li><a className="dropdown-item" href={env.logoutURL} onClick={() => user.logout()}>Log out</a></li>
                             </Menu>
```
nathanstitt, line 28: I think this will setup a race between the navigation to `logoutURL`  cancelling any running JS, and the logout() call sending the fetch request to the backend.  To work around that you can make the logout method be bound to the user instance and call it directly here: `onClick={user.logout}`

then have the user method it look like:
```ts
  logout: (ev?: React.MouseEvent<HTMLAnchorElement>) => { 
   if (!ENV.IS_DEV) return

   ev?.preventDefault() 
   .. rest of method
 }
```

frontend/src/models/environment.ts
```diff
@@ -1,6 +1,5 @@
 import { API_CONFIGURATION, ENV } from '@lib'
-import { Environment as ApiEnv, DefaultApi } from '@api'
-import dayjs from 'dayjs'
```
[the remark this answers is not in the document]
chrisbendel: interesting indeed!!

PR #190, opened 2022-09-28
LD style/layout tweaks

chrisbendel:


frontend/src/components/footer.tsx
```diff
@@ -46,34 +51,72 @@ const Funders = () => {
...
+                                    </a>
+                                </Box>
+                            </Row>
+                            <Col className="py-1" direction="row" justify="around">
+                                <div><img alt="Rice University logo" height="30" src={RiceLogoURL} /></div>
+                                <div><img alt="Open Stax logo" height="30" src={OpenStaxURL} /></div>
+                            </Col>
+                        </Row>
+                    </div>
                 </div>
```
nathanstitt, line 118: yeah that's kinda hard to read 😁   How about extracting each side of the ternary out into separate components and then doing something like: 
```ts
export const Footer: React.FC<{ className?: string, includeFunders?: boolean }> = ({ className, includeFunders }) => (
  <div className={cx('footer', 'mt-4', className)}>
      {includeFunders && < Funders />}
      {useIsMobileDevice() ? <MobileFooter />  : <DesktopFooter />}
</div>
)
```

nathanstitt (changes requested):
looks good other than the ternary could be extracted

frontend/src/styles/main.scss
```diff
@@ -6,8 +6,10 @@ $link-color: #0d6efd;
 
 @import "./bootstrap";
 @import "flatpickr/dist/flatpickr.css";
-@import "./effects";
 
+#root {
+    overflow: hidden;
+}
```
nathanstitt, line 12: Hm, this is a pretty big rule to bust out here.  I'd prefer to figure out what's overflowing and fix that rather than hiding it.

I'm assuming this is for the prevent side-scroll card?  I'd played with it on iOS but couldn't see any overflow,  I do have an older android phone for testing though and was going to attempt that.  I can do so and let you know what I find out

nathanstitt (approved):
Looks good other than the overflow rule, let me test that and we'll chat before merging

PR #192, opened 2022-10-04
construct "personality" icon using '#' char overlay

nathanstitt:
This is how the figma did it, which meant there wasn't an straight SVG to download.  this seems to work well though

fixes https://github.com/openstax/research/issues/205

image


chrisbendel (approved):
I'm surprised viki didn't choose one from the available icons, but looks good to me otherwise

nathanstitt:
yeah, I tried to find one, but there doesn't appear to be a match for a # chat icon: https://icons.getbootstrap.com

PR #195, opened 2022-10-04
Fix homepage mobile supporter section, align with styles on LD

chrisbendel:
fixes openstax/research#198


nathanstitt (approved):
looks great Chris, thanks

PR #196, opened 2022-10-05
Use flex box to layout segments

nathanstitt:
This allows the height to be set based on needed label height, whereas absolute positioning would overflow if longer than 2 lines

fixes https://github.com/openstax/research/issues/202

### DIFF
--- a/frontend/src/components/rewards-progress-bar.tsx
+++ b/frontend/src/components/rewards-progress-bar.tsx
@@ -75,7 +75,7 @@ const GrandPrize: React.FC<{ segment?: RewardsSegment }> = ({ segment }) => {
     if (!segment) return null
 
     if (segment.isPast && !segment.achieved) {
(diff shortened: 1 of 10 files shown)


frontend/src/components/rewards-progress-bar.tsx
```diff
@@ -75,7 +75,7 @@ const GrandPrize: React.FC<{ segment?: RewardsSegment }> = ({ segment }) => {
     if (!segment) return null
 
     if (segment.isPast && !segment.achieved) {
-        return <MissedGrandPrize css={{ margin: '-10px 0 10px 50px' }} />
+        return <MissedGrandPrize css={{}} />
```
chrisbendel, line 78: do we need empty css attribute?

chrisbendel (approved):

chrisbendel (approved):
Looks good!

PR #200, opened 2022-10-13
215 study card responsiveness

chrisbendel:


frontend/src/screens/learner/card.tsx
```diff
@@ -128,19 +143,28 @@ export const StudyCard: React.FC<StudyCardProps & { onSelect(study: ParticipantS
             data-is-completed={!!study.completedAt}
             onClick={onClick}
         >
-            <Image.image name={Image.title} height="200px" css={{ border: `1px solid ${colors.lightGray}`, borderRadius: 8 }} />
+            <Image.image
+                name={Image.title}
+                height={useIsMobileDevice() ? '97px' : '200px'}
```
nathanstitt, line 148: instead of using isMobile, we could use media.mobile in the css.  Also 97px seems oddly specific, guessing that was needed for something, but it makes me nervous that things will overflow on different browsers 🤔

frontend/src/styles/main.scss
```diff
@@ -6,10 +6,21 @@ $link-color: #0d6efd;
 
 @import "./bootstrap";
 @import "flatpickr/dist/flatpickr.css";
+@import 'swiper/swiper.min.css';
+@import 'swiper/swiper-bundle.css';
```
nathanstitt, line 10: if swiper provides a non-minimized css it'd be better to use it because then it'd be legible in our source maps.   We compress the css ourselves when bundling and doing it twice is extra work

frontend/src/styles/main.scss
```diff
@@ -6,10 +6,21 @@ $link-color: #0d6efd;
...
 }
+
+.truncate-2 {
+    overflow: hidden;
+    text-overflow: ellipsis;
+    display: -webkit-box;
+    -webkit-line-clamp: 2;
+    line-clamp: 2;
+    -webkit-box-orient: vertical;
+}
```
nathanstitt, line 26: neat! I wasn't aware of `line-clamp`

frontend/src/components/icon.tsx
```diff
@@ -40,12 +40,10 @@ export const Icon = React.forwardRef<SVGSVGElement, React.PropsWithChildren<Icon
     forwardedProps,
     ref
 ) => {
-
     const {
         id, icon: iconProp, tooltip, onClick, children, busy, buttonStyles = {}, tooltipProps = {}, ...iconProps
     } = forwardedProps
 
-    //  ) => {
```
nathanstitt, line 48: commented code can be removed

nathanstitt (approved):
looks great Chris! just a few suggestions but nothing major

frontend/src/screens/learner.tsx
```diff
@@ -61,6 +57,37 @@ const StudyList: FCWOC<StudyListProps> = ({ className, onSelect, title, studies,
...
 
+const MobileStudyList: FCWOC<StudyListProps> = ({ className, onSelect, title, studies, children }) => {
+    return (
+        <div className={cx('container-lg', 'studies', 'my-4', className)}>
+            <h3 className='py-2'>{title}</h3>
+            {children}
+            {!studies.length && <h3>Awesome, you completed all studies! Watch out for new studies coming up soon!</h3>}
+
+            {chunk(studies, 6).map((studyChunk, i) => {
+                return (
```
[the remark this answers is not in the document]
chrisbendel: aah yes, leftover from when i was experimenting with bootstraps carousel and assigning something :) good catch

frontend/src/styles/main.scss
chrisbendel: me either 😆

frontend/src/screens/learner/card.tsx
chrisbendel: yeah 97px came from the figma design... we can iterate on this with iris/viki down the line if needed

chrisbendel:
I'm going to wait on any merging until viki/iris make any final design decisions on the swiper component, once that's good we can go ahead with this

PR #204, opened 2022-10-24
Add new codes for staging/prod

chrisbendel:


nathanstitt:
Hey @chrisbendel, this was already completed and I failed to move the card onto done, sorry!  I did it by just setting the heroku env `OX_GA_UA` to the appropriate values on staging and prod.  if you run `heroku config:get GA_UA --app kinetic-web` you should see the new values

PR #207, opened 2022-11-09
204 study session points time

chrisbendel:


.gitignore
```diff
@@ -2,6 +2,7 @@
 tmp
 .DS_Store
 docker-compose.override.yml
+docker-compose.dev.yml
```
chrisbendel, line 5: i added a local docker override to run pgadmin in locally to mess around with the database GUI :)

nathanstitt: yeah that's good, I'd added the line 4 earlier, my override is `docker-compose.override.yml` previously. How about we make the file `docker-compose.override.dev.yml` or something so that it's clear that it's an override file that's for local dev?

frontend/src/components/icon.tsx
```diff
@@ -66,6 +76,7 @@ export const Icon = React.forwardRef<SVGSVGElement, React.PropsWithChildren<Icon
             <button
                 onClick={onClick}
                 disabled={busy}
+                data-test-id={dti}
                 css={Object.assign({
                     border: 'none',
                     padding: 0,
```
nathanstitt, line 82: you'll also need to set the `data-test-id` on the icon if the button isn't rendered, otherwise it'll be lost

frontend/src/screens/learner.tsx
```diff
@@ -51,7 +51,7 @@ const StudyList: FCWOC<StudyListProps> = ({ className, onSelect, title, studies,
             {children}
             {!studies.length && <h3>Awesome, you completed all studies! Watch out for new studies coming up soon!</h3>}
             <Grid css={{ overflow: 'auto', paddingBottom: '10px' }} data-test-id="studies-listing">
-                {studies.map(s => <StudyCard onSelect={onSelect} study={s} key={s.id} />)}
+                {studies.map((s, i) => <StudyCard onSelect={onSelect} study={s} key={i} />)}
             </Grid>
```
nathanstitt, line 55: if possible, it's preferable to use something other than the index for a key. https://articles.wesionary.team/using-index-as-a-key-is-an-anti-pattern-in-react-8e5db3aea3a9

nathanstitt (approved):
Looking great Chris, just a few comments about using the index as `key`, also the FE specs look like they timed-out.

frontend/specs/helpers.ts
```diff
@@ -75,6 +75,7 @@ export const logout = async ({ page }: { page: Page }) => {
 export const loginAs = async ({ page, login }: { page: Page, login: TestingLogin }) => {
     await logout({ page })
     await page.goto('http://localhost:4000/dev/user')
+    await page.waitForSelector('.dev-console');
```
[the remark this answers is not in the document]
chrisbendel: there was an odd issue i was experiencing here, as well as a couple of other pages (where i had added a timeout). the page would have a "Loading..." for a brief second, and these seemed to be running into issues on those pages

PR #208, opened 2022-11-10
Fix banner and splash image on mobile

chrisbendel:


nathanstitt (approved):
lgtm other than checkout the empty div thing

PR #212, opened 2022-12-01
Generate test data for a study

nathanstitt:
Sets up a framework for downloading Qualtrics data


backend/app/controllers/api/v1/researcher/study_responses_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def fetch
+    @study.fetch_responses(is_testing: @is_testing)
+    show
+  end
+
+  protected
+
+  def set_study
+    # TODO: set is_testing only when coming from outside network
```
chrisbendel, line 25: If we're doing this in the future let's make a ticket so we don't lose sight

backend/config/application.rb
```diff
@@ -21,7 +21,7 @@
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
-Dotenv.load('/etc/.env', '.env')
+Dotenv.load('/etc/.env', '.env', '.env.local')
```
chrisbendel, line 24: Is .env.local something we'll have to set up locally that will be ignored in source control? if so just making sure i don't lose sight of this :D

backend/app/models/study_response_export.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  has_many_attached :files
+
+  def fresh?
+    is_complete && updated_at.after?(4.hours.ago)
+  end
+
+  def fetch
+    return if is_complete?
+    # TODO: kickoff and monitor Qualtrics download
```
chrisbendel, line 22: lets create a followup ticket before/when we merge

chrisbendel (approved):

backend/app/models/study_response_export.rb
nathanstitt: https://github.com/openstax/research/issues/318

PR #221, opened 2023-01-13
generate enclave data 

nathanstitt:
Adds an "analysis" model and methods to generate fake data around it


chrisbendel:
Noticed a couple small things in the test results:

1) the ruby version seems to be mismatched somehow?

2) something is up with the minutes/points changes in #207 , should we merge that PR first?

backend/app/services/qualtrics_test_data.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'faker'
```
chrisbendel, line 3: is faker used here?

chrisbendel (approved):

backend/app/services/qualtrics_test_data.rb
nathanstitt: yes, it's the calls to `Faker:Lorem` below

chrisbendel:
Looks good to me!!

PR #235, opened 2023-02-14
Researcher studies dashboard

chrisbendel:


backend/app/controllers/api/v1/researcher/studies_controller.rb
```diff
@@ -42,6 +42,7 @@ def update
   def destroy
     @study.update!(is_hidden: true)
     head :ok
+    # render json: @study, status: :ok
```
chrisbendel, line 45: should we return the study on delete, since it's not a hard delete?

PR #237, opened 2023-03-08
A few small changes to support the workspaces editor

nathanstitt:


bin/update-bindings
```diff
@@ -4,6 +4,24 @@ set -xe
 
 cd backend
 
+while getopts ":h:upload:" o; do
```
chrisbendel, line 7: this looks like dark magic to me 😆

chrisbendel (approved):

PR #247, opened 2023-05-31
API updates to workspaces testing data

nathanstitt:
Adds an admin panel for explicitly setting study data and replies with that if it's available before generating it's own random data


backend/spec/requests/api/v1/researcher/studies_spec.rb
chrisbendel, line 1: some odd formatting in this file

backend/lib/tasks/populate_study_researcher_data.rake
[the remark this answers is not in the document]
nathanstitt: oh did you add that?  rubocop didn't like an empty file so I'd removed it

backend/config/secrets.yml
```diff
@@ -5,6 +5,7 @@ development:
   qualtrics_api_key: 000000
   qualtrics_api_url: https://iad1.qualtrics.com/API/v3
   qualtrics_group_library: 'GR_7R5NeLAPtj22K46'
+  enclave_api_key: ec_1bG346BDBGEsG5Dck8H6sAiRBJ7Vt52Vwcno
```
chrisbendel, line 8: should these api keys be env-ified?

nathanstitt: it isn't a "real" api key, it's just for local testing, note that it's only in the dev & test blocks.  

in production it does read from ENV

frontend/src/screens/admin/grid.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { BoxProps } from 'boxible'
+
+export const Grid = styled.div({
+    display: 'grid',
+    height: '100vh',
+    width: '100vw',
+    gridTemplateColumns: 'minmax(auto, 200px) 1fr',
+    gridTemplateRows: 'auto 1fr auto',
+    overflow: 'hidden',
+    gridTemplateAreas: `
```
chrisbendel, line 11: This is super cool, i didn't know this was a thing. read up on it a little more and wonder if down the line we could build out the whole site's structure like this!

backend/lib/tasks/populate_study_researcher_data.rake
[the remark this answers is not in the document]
chrisbendel: actually i dont think we'll need it, good to keep it out! 👍

PR #238, opened 2023-03-15
Researcher Dashboard Milestone

chrisbendel:

PR #249, opened 2023-06-21
enclave data

nathanstitt:


backend/spec/factories/analyis_runs.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :analyis_run do
+
```
chrisbendel, line 3: need anything in here?

.gitignore
```diff
@@ -5,5 +5,6 @@ docker-compose.override.yml
 docker-compose.override.dev.yml
 backend/clients
 frontend/test-results
+backend/data
```
chrisbendel, line 8: would I need any local data to make things happen since this is gitignored? or is this just a place for local data to live once it's generated?

backend/spec/requests/api/v1/enclave_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        expect {
+          api_post 'enclave/runs/log', params: { api_key: run.api_key, stage: 'review', level: 'info', message: 'some code to test' }
+        }.to change { AnalysisRunMessage.count }.by(1)
+      end
+
+      it 'sends a completion notice' do
+        run.messages.create({ stage: 'archive', level: 'info', message: 'something' })
+        expect {
+          api_put 'enclave/runs/completion', params: { api_key: run.api_key, status: 'failure', error: 'a bad thing happened' }
+          expect(run.messages.last.message).to eq 'a bad thing happened'
```
chrisbendel, line 48: nomination for best error message of the year

backend/config/routes.rb
```diff
@@ -18,7 +18,9 @@
         end
         post 'studies/:id/update_status', to: 'studies#update_status'
 
-        resources :analysis, except: [:destroy]
+        resources :analysis, except: [:destroy] #  do
+        #   resources :runs, only: [:create], shallow: true
```
chrisbendel, line 22: todo

backend/app/services/qualtrics_api.rb
```diff
@@ -21,6 +24,44 @@ def share_survey(survey_id, recipient_id, permissions)
...
+    Tempfile.create([survey_id, '.zip'], binmode: true) do |f|
+      resp.each { |chunk| f.write(chunk) }
+      f.flush
+
+      Zip::File.open(f) do |zip|
+        zip.each(&block)
+      end
+
+    end
+    #   end
```
chrisbendel, line 59: todo

backend/app/models/study.rb
```diff
@@ -50,7 +50,7 @@ class Study < ApplicationRecord
   }
 
   def status
-    # todo go off of last stage or first stage? probably last stage
+    # TODO: go off of last stage or first stage? probably last stage
```
chrisbendel, line 53: this is my mess to clean up in my PR :D

backend/app/models/response_export.rb
```diff
@@ -12,20 +12,72 @@ def self.new_random_seed
...
+  def fetch_real_responses
+    # make sure we're not already fetching
+    api = QualtricsApi.new
+    latest = stage.response_exports.where.not(id: id).real.completed.by_date.last
+
+    progress_id = api.start_response_export(
+      stage.config['survey_id'], latest&.cutoff_at, cutoff_at
+    )
+
+    # TODO: monitor this and move it to a background job if it takes more than a few seconds
```
chrisbendel, line 39: This reminds me to ask you @nathanstitt , in my PR [here](https://github.com/openstax/kinetic/pull/238/files#diff-8a40967b7f6b2a2d4777f9b2d385020f5fd82b1eb6c1eb39297e73e054ef0095R112) the qualtrics cloning takes roughly a few seconds on the request. i wonder if this is also worth doing async? probably fine for now, but figured i'd mention it in case we have a more generic way to do that with your note here

chrisbendel (approved):

.gitignore
nathanstitt: actually I'm not sure if this is needed or not, my local had some CSV's in it, which I think were a temp files from qualtrics download

backend/config/routes.rb
nathanstitt: hm, I think that can actually be removed.  I'll do so.

backend/app/models/response_export.rb
nathanstitt: yes, we should do so then.  I think we can use background job in a simple thread

PR #250, opened 2023-06-28
Fetch responses, send email

nathanstitt:


backend/lib/tasks/install_secrets.rake
chrisbendel, line 1: Just for my own knowledge - is this unused/irrelevant or is this happening elsewhere?

chrisbendel (approved):

PR #255, opened 2023-07-10
Study dashboard hotfixes

chrisbendel:


backend/app/controllers/api/v1/researcher/studies_controller.rb
```diff
@@ -37,9 +37,9 @@ def update
     render(json: error, status: error.status_code) and return if error
 
     notify_researchers(inbound_binding.researchers || [])
-    # TODO: after fleshed out instructions
-    # @study.reopen_if_possible(inbound_binding.to_hash)
+
     @study.update(inbound_binding.to_hash.except(:researchers, :stages))
```
nathanstitt, line 41: This must have snuck through earlier review, but it needs to be either the "bang" form as `update!` that will throw an exception, or need to check the error return to make sure the update succeeded

nathanstitt (approved):
looks good, we can fix the update error check here or later in a different PR if you'd like

PR #256, opened 2023-07-20
Bootstrap cleanup theme

chrisbendel:


nathanstitt (approved):
looks great other than the sundryIcons need to be set still.  thanks Chris

PR #266, opened 2023-07-31
Analysis tutorial overlay

chrisbendel:


frontend/src/screens/analysis/analysis-tutorial.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const renderStep = (step: number, setStep: (step: number) => void) => {
+        switch(step) {
+            case 0:
+                return <Step1 step={step} setStep={setStep} />
+            case 1:
+                return <Step2 step={step} setStep={setStep} />
+            case 2:
+                return <Step3 step={step} setStep={setStep} close={closeTutorial} />
+        }
+    }
```
nathanstitt, line 30: This doesn't appear to rely on any closure variables, so why not move it out of the AnalysisTutorial function and make it a "real" component `TutorialStep` or something, then you can render it below as `<TutorialStep index={step} onChange={stetStep} />`

nathanstitt (approved):
looks good, just a small comment around the step component

PR #268, opened 2023-08-07
Public on migration with updated new data

chrisbendel:


backend/db/migrate/20230404161002_exports_cutoff_at.rb
```diff
@@ -1,5 +1,14 @@
...
+    add_column :studies, :public_on, :datetime, null: true
     add_column :analysis_response_exports, :cutoff_at, :datetime
+
+  end
+
+  def down
+    remove_column :studies, :public_on, :datetime
+    add_column :studies, :shareable_after_months, :integer
+    remove_column :analysis_response_exports, :cutoff_at, :datetime
   end
```
nathanstitt, line 13: no reason to do `up` & `down` migrations, add_column and remove_column are both reversible and only require `change`

nathanstitt (approved):

backend/db/migrate/20230404161002_exports_cutoff_at.rb
chrisbendel: got it - thanks! copy/paste syndrome got me

PR #270, opened 2023-08-08
Use Form instead of EditingForm

chrisbendel:


nathanstitt (approved):

backend/app/models/launched_study.rb
```diff
@@ -27,15 +25,15 @@ def opted_out!
...
   def consented!
     update!(consent_granted: true)
   end
 
   def completed!
     update!(completed_at: Time.now)
     completed = LaunchedStudy.where(study_id: study_id).complete.count
-    Study.update_counters(study_id, completed_count: completed)
+    Study.find(study_id).update_columns(completed_count: completed)
   end
```
nathanstitt, line 37: I didn't think this one through when I'd suggested `update_columns` and didn't realize that we weren't already loading study.  Doing it this way requires fetching study from the DB, parsing it into a ruby class instance, then updating the cols.  it's better to just do the raw sql using `update_all` https://api.rubyonrails.org/v3.2.3/classes/ActiveRecord/Relation.html#method-i-update_all
should work I think: 
`Study.update_all({ completed_count: completed}, { id: study_id })`

nathanstitt (approved):

PR #271, opened 2023-08-15
Admin impersonation

chrisbendel:


frontend/src/screens/dev/users.ts
```diff
@@ -24,6 +29,7 @@ export class AvailableUsers {
     researchers: User[] = []
 
     constructor({ admins, users, researchers }: UsersPayload = { admins: [], users: [], researchers: [] }) {
+        console.log(admins, users, researchers)
```
nathanstitt: remove console.log

frontend/src/lib/environment-provider.tsx
```diff
@@ -1,55 +1,72 @@
...
+export const useCurrentUser = () => {
+    const env = useEnvironment()
+    return env?.user
+}
 
-export const useCurrentResearcher = () => useEnvironment()?.researcher
+export const useCurrentResearcher = () => {
+    const env = useEnvironment()
+    return env?.researcher
+}
```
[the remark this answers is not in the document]
chrisbendel: got it - i see how you were doing it earlier. We don't load the main app until env is populated anyway, the key was `as Environment` on the previous line 40 :D

frontend/src/main.tsx
```diff
@@ -21,18 +21,18 @@ const queryClient = new QueryClient({
...
+            <QueryClientProvider client={queryClient}>
                 <EnvironmentProvider>
-                    <QueryClientProvider client={queryClient}>
+                    <MantineProvider withGlobalStyles withNormalizeCSS theme={mantineTheme}>
                         <AppRoutes />
-                    </QueryClientProvider>
+                    </MantineProvider>
                 </EnvironmentProvider>
-            </ThemeProvider>
+            </QueryClientProvider>
```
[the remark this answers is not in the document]
chrisbendel: This should be all good - it ends up actually fixing a font issue viki and I noticed :D

backend/app/bindings/api/v1/bindings/dev_user.rb
nathanstitt, line 1: I don't think we want to document the dev user. if you're only doing so for the type definitions, we can just write an interface

backend/app/controllers/api/v1/dev_open_api.rb
nathanstitt, line 1: same as above, I'd prefer to leave this undocumented unless there's a benefit to doing so

frontend/src/components/resource-links.tsx
```diff
@@ -12,7 +12,7 @@ export const ResourceLinks = () => {
 }
 
 export const HelpLink = () => {
-    const isResearcher = useCurrentUser().isResearcher
+    const isResearcher = useCurrentUser()?.isResearcher
```
nathanstitt, line 15: can `useCurrentUser()` return an empty object while its loading?  that way you wouldn't need to remember to add a `?` everywhere

nathanstitt (commented):
looking good, except I'd prefer to remove the openapi definitions for dev methods, and also the admin links load only for admin

frontend/src/components/navbar/top-navbar.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <Menu shadow='md' opened={opened} onChange={setOpened} width={175}>
+            <Menu.Target>
+                {menuToggle}
+            </Menu.Target>
+
+            <Menu.Dropdown>
+                {isMobile && <Menu.Item><StyledLink to="/studies">Studies</StyledLink></Menu.Item>}
+                <AdminLinks />
+                <AccountLinks />
+            </Menu.Dropdown>
```
nathanstitt, line 76: hah, sorry I wasn't fully clear on how the async works.  it loads the component whenever it's rendered so this usage will force it to load all the time, defeating the purpose. You'll need to change this to be something like:
```
   {user.isAdmin && <AdminLinks />}
```

Alternatively we could extend the `loadAsync` function to have a `when` argument that it would call to determine if it should render or not,.

frontend/src/components/navbar/top-navbar.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <nav className={cx('navbar', 'navbar-light', className)}>
+            <div className="navbar-dark bg-dark py-1">
+                <div className="container-lg">
+                    <Box justify="between" align="center" gap padding={{ vertical: 'default' }}>
+                        <NavbarLogoLink />
+                        {children}
+                        <Box gap="xlarge" align="center">
+                            {!isMobile && user.isResearcher && <Link to="/studies" css={{ color: 'white', textDecoration: 'none', fontWeight: 'bold' }}>Studies</Link>}
+                            {!isMobile && user.isResearcher && <Link to="/analysis" css={{ color: 'white', textDecoration: 'none', fontWeight: 'bold' }}>Analysis</Link>}
+                            <NavMenu />
```
nathanstitt, line 40: I know this isn't new code, but you could make a `DesktopResearcherLink` to that returns null when the conditions aren't met

frontend/src/lib/environment-provider.tsx
```diff
@@ -59,3 +70,39 @@ export const useUserPreferences = () => {
...
     })
 }
+
+export const host = () => {
+    const env = useEnvironment()
+    if (env.accountsEnvName === 'production') {
+        return `https://openstax.org`;
+    }
+    return `https://${env.accountsEnvName}.openstax.org`;
+}
```
nathanstitt, line 80: this is actually returning the `origin`, host lacks the scheme.  maybe rename it to be `locationOrigin`?

frontend/src/screens/dev/index.tsx
```diff
@@ -46,23 +49,16 @@ export default function Dev() {
         const userId = ev.currentTarget.dataset.userId
         ev.preventDefault()
         if (userId) {
-            await currentUser.become(userId)
+            await loginAsUser(userId)
+            await refetchEnvironment()
             nav('/')
```
nathanstitt, line 54: `refetchEnvironment()` doesn't actually do anything here, right?  they browser is navigating away from the page so the request will probably be canceled as it does, and the result will never be utilized.

frontend/src/models/user.ts
```diff
@@ -66,6 +58,42 @@ export class User {
...
+    await fetch(`${ENV.API_ADDRESS}/development/users/log_out`, {
+        method: 'DELETE', credentials: 'include',
+    })
+}
+
+export const loginAsUser = async (id: string) => {
+    await fetch(`${ENV.API_ADDRESS}/development/users/${id}/log_in`, {
+        method: 'PUT', credentials: 'include',
+    })
+}
```
nathanstitt, line 72: this should be using `window.location = <href>` not fetch, OR change the caller to just use an `<a href={`${ENV.API_ADDRESS}/development/users/${id}/log_in`}>Become User</a>` tag.

nathanstitt (commented):
need to switch the impersonation call to a link

frontend/src/lib/environment-provider.tsx
chrisbendel: oh yeah this was just moved so i didn't actually 'change' anything here, happy to change this to be more clear though :D

frontend/src/screens/dev/index.tsx
chrisbendel: interestingly enough without refetchEnvironment we still have that issue where you can
- login as researcher
- logout
- login as user
- redirect to '/'
- you still see the researcher view

chrisbendel: i think refetch would be forcing the environment provider to update the state and propagate that down through the context

frontend/src/models/user.ts
nathanstitt: oh no sorry! I got confused and thought this was the impersonate call 😭   Sorry to confuse ya, this is fine since it's dev-only, although maybe someday we should update it as well

nathanstitt (approved):

PR #276, opened 2023-08-17
add separation between researcher-bio + logic for lead = pi

raissinging:
[Study Card - About Researcher #484](https://app.zenhub.com/workspaces/kinetic-611c260d2f76a4001cd83a6f/issues/gh/openstax/research/484)

### DIFF
--- a/frontend/src/screens/learner/details.tsx
+++ b/frontend/src/screens/learner/details.tsx
@@ -185,7 +185,7 @@ export const StudyDetails: React.FC<{ studies: ParticipantStudy[] }> = ({ studie
     )
 }
 
-const ResearcherSection = (study: Study | ParticipantStudy) => {
+const ResearcherSection: FC<StudyDetailsProps> = ({ study }) =>  {
     const pi = getStudyPi(study)
     const lead = getStudyLead(study)
 
@@ -229,7 +229,7 @@ export const StudyDetailsPreview: FC<{
                     <StudyPart property="feedbackDescription" title="Feedback Available" icon="feedback" study={study} />
                     <MultiSession study={study} />
                     <Box margin={{ bottom: 'large' }} css={{ color: colors.text }}>{study.longDescription}</Box>
-                    {renderResearcherSection(study)}
+                    <ResearcherSection study={study} />
                     <StudyPart property="benefits" title="What’s in it for you" icon="heart" study={study} />
                     <Part icon="warning" title="Notice">
                         Your responses to this study will be used to further learning science and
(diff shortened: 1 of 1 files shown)


frontend/src/screens/learner/details.tsx
```diff
@@ -208,8 +215,16 @@ export const StudyDetailsPreview: FC<{
                     <StudyPart property="feedbackDescription" title="Feedback Available" icon="feedback" study={study} />
                     <MultiSession study={study} />
                     <Box margin={{ bottom: 'large' }} css={{ color: colors.text }}>{study.longDescription}</Box>
-                    {pi && <Researcher researcher={pi} />}
-                    {lead && <Researcher researcher={lead} />}
+                    {
```
chrisbendel, line 218: Good job! Just a little clean-up comment

Since the logic is getting a little more complex here (inside the render function), we can probably move this researcher section to it's own component. You can still keep the old <Researcher> component, but maybe add a <ResearcherSection> which you can pass the study into. 
  
Then you can extract all this logic out into that component and keep this one a little cleaner 😄 .

Once its in its own component, you can just do something like 
```
if (pi && lead && pi.id === lead.id) P
  return <Researcher researcher={pi} />
}

return (
  <>
    {pi && <Researcher researcher={pi} />}
    {lead && <Researcher researcher={lead} />}
  </>
)
```
something like that 🤔 just the general idea if that makes sense!

chrisbendel (changes requested):

frontend/src/screens/learner/details.tsx
raissinging: Sounds good, makes sense! I implemented those changes in the commit below. Not sure if I did it correctly, so please let me know if there is anything else I need to fix. Thanks!

frontend/src/screens/learner/details.tsx
```diff
@@ -178,14 +185,28 @@ export const StudyDetails: React.FC<{ studies: ParticipantStudy[] }> = ({ studie
     )
 }
 
+const renderResearcherSection = (study: Study | ParticipantStudy) => {
```
chrisbendel, line 188: Good job! Only comment here is instead of having this be a function and calling it in the render method on line 232, you can just make this a "component" like so:

```suggestion
const ResearcherSection = (study: Study | ParticipantStudy) => {
```

and then when using this:
```
<s>{renderResearcherSection(study)}</s>
<ResearcherSection study={study} />
```

chrisbendel (approved):

PR #277, opened 2023-08-18
Generate report with admin frontend

chrisbendel:


backend/app/controllers/api/v1/admin/reports_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class Api::V1::Admin::ReportsController < Api::V1::Admin::BaseController
+  def learner_activity
+    months_ago = params[:months_ago] || 1
+    csv = LearnerActivityReport.new(months_ago: months_ago)
```
nathanstitt, line 6: this is trifling, but the `csv` variable isn't holding a CSV, it's holding the `report`, so it should be renamed something accordingly, maybe `report` or `lar` like we did below

backend/app/services/learner_activity_report.rb
```diff
@@ -0,0 +1,85 @@
...
+  def initialize(months_ago:)
+    @months_ago = months_ago
+    @user_uuids = []
+    @csv_string = CSV.generate do |csv|
+      generate_report(csv)
+    end
+  end
+
+  def as_csv_string
+    @csv_string
```
nathanstitt, line 15: this just occurred to me but it might be better to move the block from initialize to here so it's only called when a CSV string is requested.
```
def as_csv_string
    CSV.generate do |csv|
       generate_report(csv)
     end
end
```

nathanstitt (approved):
looks good just a few minor suggestions

PR #279, opened 2023-08-22
Ies nsf homepage

chrisbendel:
Waiting on iris/debshila for some decisions. saving as a draft PR for now


nathanstitt (commented):
The images you removed from homepage weren't used directly but were kept around so we had the originals in case we needed to modify or re-upload them.  Is probably fine to remove though since Im sure UX has them and I think we can always redownload the original from the CMS

chrisbendel:
Closing in favor of #293 where other related changes were made

PR #287, opened 2023-08-31
Allow for adding a custom subject

chrisbendel:


nathanstitt (approved):
What does this do to the filters for the learners? Does it appear in the "Others" section?

chrisbendel:
This should only affect the "Subject" not the "Topic", so should be all good 👍

PR #288, opened 2023-08-31
Readme update

raissinging:
Updated readme to contain the localhost link instead of the old link. I'm not too familiar with writing readmes/documentation in general so please let me know if there is anything I need to change or add/delete. Thanks!


README.md
```diff
@@ -1,5 +1,29 @@
...
+$> bin/rails db:migrate RAILS_ENV=development
+```
+
+and try running the docker compose commands again to see if that fixes it.
+
+When run correctly the front-end may be at http://localhost:4000/.
+
 ## Frontend
 
 To run the front-end for development:
```
chrisbendel, line 29: We can probably create a separate "README.md" file in the `frontend` directory and keep some of the frontend specific stuff like this in there

This section tripped my up when I first started, as I had assumed we needed the frontend running in a separate terminal. Since the docker container runs the FE for us, this is mostly irrelevant (I am trying to think of a situation where we'd run the FE without the BE)

chrisbendel (approved):

PR #289, opened 2023-09-01
Review study tweaks

chrisbendel:
Covers:
[Review study updates](https://app.zenhub.com/workspaces/kinetic-611c260d2f76a4001cd83a6f/issues/gh/openstax/research/487)


frontend/src/screens/researcher/studies/create/forms/review-study.tsx
```diff
@@ -186,14 +191,15 @@ export const SubmitStudyModal: FC<{
...
+            closeBtn={false}
             show={show}
             large
         >
             <Modal.Body>
                 <Box padding='4rem' align='center' justify='center' direction='column' gap='xlarge'>
                     <Box align='center' className='text-center' direction='column'>
-                        <p>You’re about to submit your study to the Kinetic team so that the appropriate permissions are set. Please review and confirm any final changes.</p>
-                        <p className='fw-bold text-danger'>You won’t be able to change your Kinetic study information past this point.</p>
+                        <p>You're about to submit your study to the Kinetic team so that the appropriate permissions are set. Please review and confirm any final changes.</p>
```
nathanstitt, line 201: personally I prefer the utf apostrophe bc it slants nicer 😁

nathanstitt (approved):

frontend/src/screens/researcher/studies/create/forms/review-study.tsx
chrisbendel: hah - this was driving me mad because i was searching my repository using the `'` character, and i thought i was losing my mind because no results were turning up!

Maybe we should set some standard in the repo? I know this is a minor thing but between the ` (backtick), `'` and `"`perhaps this is something the linter could enforce? I usually prefer `'` for strings, but if there are strings with a `'` character in the string, ill defer to using `"`

i know this is splitting hairs, but just figured id mention it 😆

chrisbendel: @nathanstitt i just checked our `eslintrc.js` file and we have this entry in here:
```'quotes': [ 'error', 'single', { avoidEscape: true, 'allowTemplateLiterals': true } ],```

I wonder what you think about using double quotes or backticks as the default and disallowing single quotes? [Docs for reference](https://eslint.org/docs/latest/rules/quotes#options)

PR #290, opened 2023-09-01
Fix namespace for reports open api

chrisbendel:


nathanstitt (approved):
sorry it took so long!

nathanstitt (approved):

PR #296, opened 2023-09-21
Study creation small fixes

raissinging:
Completed items in this [ticket](https://app.zenhub.com/workspaces/kinetic-611c260d2f76a4001cd83a6f/issues/gh/openstax/research/488)


frontend/src/screens/researcher/studies/overview/finalize-study.tsx
```diff
@@ -82,21 +83,103 @@ const QualtricsConfirmationContainer: FC<{
...
+        <div>
+            <h6
+                style={{
+                    textDecoration: 'underline',
+                    textUnderlineOffset: '.5rem',
+                    cursor: 'pointer',
+                }}
+                onClick={() => {
+                    if (showWarning) {
+                        setShowWarning(false);
```
chrisbendel, line 128: Currently the warning never gets triggered by the form, this is an inherent problem with the form context only being available "inside" the form. since this button is outside, there doesn't seem to be an obvious way to get the forms state (`dirty` in this case)

I wonder if we can talk to iris/viki and change the approach. If we could get away with a `Window.confirm()` and utilize the default browser rather than rolling our own. That would solve this problem in a couple different areas

@vikizhu5504 @iccgoncalves

chrisbendel (approved):

PR #299, opened 2023-09-28
initial code to block navigation w/form dirty

nathanstitt:
CC @chrisbendel I think this is what we'll want instead of https://github.com/nathanstitt/sundry/pull/10

It optionally blocks history.pushstate in react-router using https://www.npmjs.com/package/react-router-prompt

Unfortunately to use it we need to upgrade the router to use their new "data routes". I did a first pass at it, but some of the routes may still be broken


frontend/src/lib/environment-provider.tsx
```diff
@@ -30,7 +29,7 @@ export const EnvironmentProvider: FCWC = ({ children }) => {
         return <LoadingAnimation />
     }
 
-    if (!env.user.userId && (ENV.IS_PROD_MODE || !location.pathname.startsWith('/dev/user'))) {
+    if (!env.user.userId && (ENV.IS_PROD_MODE || !window.location.pathname.startsWith('/dev/user'))) {
```
nathanstitt, line 32: this blue up because useLocation wasn't being called inside a route provider.  I guess the old version of the router was more forgiving

chrisbendel (approved):

PR #305, opened 2023-11-01
Various fixes for prod

chrisbendel:


backend/app/models/study.rb
```diff
@@ -80,6 +80,16 @@ def launched_count
     launched_studies.size
   end
 
+  def update_stages(updated_stages)
+    return if updated_stages.nil? || launched_stages.any?
```
chrisbendel, line 84: Added the `launched_stages.any?` check, as we don't want to allow stages to be updated after users have launched

nathanstitt (approved):

PR #313, opened 2023-11-07
User info fetch

chrisbendel:


nathanstitt (approved):
all looks good to me.  i'm a bit unclear on where the actual bug was though

frontend/src/lib/environment-provider.tsx
```diff
@@ -70,38 +72,39 @@ export const useUserPreferences = () => {
     })
 }
 
-export const locationOrigin = () => {
+export const useLocationOrigin = () => {
     const env = useEnvironment()
```
chrisbendel, line 76: @nathanstitt pretty sure this was the line that was causing staging/prod to blow up. it may have just been a simple naming thing, since locationOrigin wasn't a hook! super interesting

PR #314, opened 2023-11-08
Various fixes

chrisbendel:


backend/app/models/launched_study.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class LaunchedStudy < ApplicationRecord
-  belongs_to :study
+  belongs_to :study, counter_cache: true
```
chrisbendel, line 4: I believe I was doing the counter_cache wrong on the has_many and it should be on the belongs_to side [according to the docs](https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache) (the values seem to be populating correctly now)

frontend/src/components/manage-cookies.tsx
```diff
@@ -15,13 +14,7 @@ export const ManageCookiesLink = () => {
         osano.cm.showDrawer('osano-cm-dom-info-dialog-open')
     }
 
-
     return (
-        <>
-            <Global
```
chrisbendel, line 21: moved to main.scss

frontend/src/components/manage-cookies.tsx
```diff
@@ -15,13 +14,7 @@ export const ManageCookiesLink = () => {
...
-                styles={css({ '.osano-cm-widget': { display: 'none' } })}
-            />
-            <a href='#' onClick={showOsano}>Manage Cookies</a>
-        </>
+        <span onClick={showOsano}>Manage Cookies</span>
```
nathanstitt, line 18: if you want to use a span, this needs to have `cursor:hover` and also a role="button" for a18y

frontend/src/screens/researcher/studies/create/action-footer.tsx
```diff
@@ -15,12 +20,12 @@ export const ActionFooter: FC<{ step: Step, }> = ({ step }) => {
...
         <Box className='fixed-bottom bg-white mt-auto' css={{ minHeight: 80, boxShadow: `0px -3px 10px rgba(219, 219, 219, 0.5)` }}>
             <Box className='container-lg' align='center' justify='between'>
-                {step.backAction ? <Link to=''>
+                {step.backAction ? <FakeLink>
                     <Box align='center' gap='small' onClick={() => step.backAction?.()}>
                         <Icon icon='chevronLeft'></Icon>
                         <span>Back</span>
                     </Box>
-                </Link> : <span></span>}
+                </FakeLink> : <span></span>}
```
nathanstitt, line 28: should add `role="link"` to the span for screenreader

frontend/src/components/manage-cookies.tsx
nathanstitt: Hm, I slightly prefer it here because this way we only hide the osano cookie link if we're also rendering an alternative method.  If the hiding is in the sass, it's possible that we hide it, but then don't render this on some screens

nathanstitt (approved):
looks good just a few accessibility things and the link

frontend/src/components/manage-cookies.tsx
chrisbendel: I should have added another comment for context woops, we moved the manage cookies link into the user name/dropdown thing at the top right, with that being there it's not technically rendered until the dropdown is opened. This way every user should have access to it (ie before the researcher experience doesn't have a footer so the cookie was still rendering)

frontend/src/components/manage-cookies.tsx
chrisbendel: @nathanstitt the cursor portion is taken care of by the parent link

```
<StyledLink to='#'>
    <Menu.Item>
        <ManageCookiesLink />
    </Menu.Item>
</StyledLink>
```
the result:

Screenshot 2023-11-13 at 3 23 45 PM

Question is, would we still want the role on the span? or would this parent cover that case?

chrisbendel: @nathanstitt the cursor portion is taken care of by the parent link

```
<StyledLink to='#'>
    <Menu.Item>
        <ManageCookiesLink />
    </Menu.Item>
</StyledLink>
```
the result:

Screenshot 2023-11-13 at 3 23 45 PM

Question is, would we still want the role on the span? or would this parent cover that case?

nathanstitt: nah, sorry I'd left that comment before I realized it was being used in the menu

PR #315, opened 2023-11-15
turn off ip spoofing checks

nathanstitt:
It seems like occasionally cloudfront will set the HTTP_X_FORWARDED_FOR to multiple IPs and trigger that check to fail.  symptoms are a ActionDispatch::RemoteIp::IpSpoofAttackError being raised resulting in a 500 error


chrisbendel (approved):

chrisbendel:
good catch!

PR #321, opened 2023-12-11
Eligibility modal

chrisbendel:


backend/app/controllers/api/v1/environment_controller.rb
```diff
@@ -14,7 +15,8 @@ def index
       homepage_url: Rails.application.secrets.homepage_url,
       banners_schedule: Banner.active.to_a || [],
       rewards_schedule: Reward.all.to_a,
-      is_impersonating: session[:impersonating].present?
+      is_impersonating: session[:impersonating].present?,
+      is_country_eligible: country.blank? || Kinetic::ELIGIBLE_COUNTRY_CODES.include?(country)
     )
```
nathanstitt, line 20: we should extract the eligibility check in the eligibility_controller to a central location so that it's identical.   maybe an services/eligibility function that you pass the request headers to?

backend/app/controllers/api/v1/environment_open_api.rb
```diff
@@ -37,6 +37,10 @@ class Api::V1::EnvironmentOpenApi
         key :type, :boolean
         key :readOnly, true
       end
+      property :is_country_eligible do
+        key :type, :boolean
+        key :readOnly, true
+      end
       property :accounts_env_name do
```
nathanstitt, line 44: I'd suggest renaming this to be "is_eligible" so the check will remain valid if we later take other things into account, such as states perhaps

nathanstitt (changes requested):

nathanstitt (approved):

PR #329, opened 2023-12-16
Only studies that have an active stage are available

nathanstitt:
Sabitha marked her study as paused but it was still showing up for participants


chrisbendel (approved):

chrisbendel:
@nathanstitt made some updates to the specs to get them passing

for the FE specs I decided to revert my approach of the `createStudiesData` raw requests and instead rely on the "user" creating studies through the UI for more robustness at the cost of a few extra seconds

PR #336, opened 2024-01-04
Add information files to a study

nathanstitt:
Intended to be HTML files that can be retrieved by analysis and displayed in the workspace


frontend/src/screens/admin/workspaces/edit-study.tsx
```diff
@@ -37,20 +37,64 @@ const Boolean:FC<{val?: boolean}> = ({ val }) => {
...
+                    <th scope="col"></th>
+                </tr>
+            </thead>
+            <tbody>
+                {infos.map(info => (
+                    <tr key={info.id}>
+                        <th>{(stages.find(s => s.id == info.stageId)?.order || 0)+ 1}</th>
+                        <th scope="row">{formatDate(info.createdAt)}</th>
+                        <td>
+                            <a href={info.url} target="_blank"><Icon icon="cloudDownload" /></a>
```
chrisbendel, line 70: I actually don't know much about the `download` attribute as I haven't used it much - mainly just commenting on this for my own self learning if you have any more info about it beyond [what's in the docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#download). Maybe just control over the filename?

chrisbendel (approved):

frontend/src/screens/admin/workspaces/edit-study.tsx
nathanstitt: the file will download due to the mime-type and content-disposition that's sent from the server

nathanstitt: however it probably wouldn't hurt to add `download`

PR #337, opened 2024-01-09
Refactor userinfo to be part of environment

chrisbendel:


backend/app/services/user_info.rb
```diff
@@ -26,7 +43,74 @@ def self.for_uuids(uuids)
...
+      {
+        user_id: admin.user_id,
+        first_name: 'Admin',
+        last_name: 'McAdminFace',
+        name: 'Admin McAdminFace',
+        isAdmin: true,
+        isResearcher: false
+      }
+    end
+  end
```
nathanstitt, line 80: a few things:
* ruby uses underscore, not camel case so it should be `is_admin` etc.
* you shouldn't load all records into memory (what if there was 1000 researchers?)  just to find a single one.  Instead make methods for the properties that are not db fields, like:



```ruby
class Researcher < ApplicationRecord
  def is_researcher
    true 
   end
```

Then later to find a given researcher, call:

`Researcher.find_by(uuid: uuid)&.to_json(methods: [:is_admin, :is_researcher])`

backend/app/services/user_info.rb
```diff
@@ -5,6 +5,23 @@
...
+      { user_id: '00000000-0000-0000-0000-000000000002', role: 'user', name: 'User Uno',
+        full_name: 'User Uno', first_name: 'User', last_name: 'Uno' },
+      { user_id: '00000000-0000-0000-0000-000000000003', role: 'user', name: 'User Dos',
+        full_name: 'User Dos', first_name: 'User', last_name: 'Dos' },
+      { user_id: '00000000-0000-0000-0000-000000000004', role: 'user', name: 'User Tres',
+        full_name: 'User Tres', first_name: 'User', last_name: 'Tres' },
+      { user_id: '00000000-0000-0000-0000-000000000005', role: 'user', name: 'User Cuatro',
+        full_name: 'User Cuatro', first_name: 'User', last_name: 'Cuatro' }
+    ].freeze
+  end
```
nathanstitt, line 23: since these are static they should be stored in a yaml file in config/ then assigned to a constant, ie:
```ruby
class UserInfo
    MOCK_USERS = YAML.load_file(Rails.root.join("config/mock-users.yaml"))
```
Also, why not make the list be a hash with uuid key so it's easier to search.   You could transform an array into hash via:
YAML.load_file(Rails.root.join("config/mock-users.yaml"))
  .reduce({}) { |h, u| h[u[:uuid]] = u; h }

backend/app/services/user_info.rb
```diff
@@ -26,7 +43,74 @@ def self.for_uuids(uuids)
...
+    if mock_user
+      u = mock_user
+    elsif researcher
+      u = researcher
+    elsif admin
+      u = admin
+    end
+
+    u
+  end
```
nathanstitt, line 97: with the above changes, I think this could be:
```
return UserInfo::MOCK_USERS[uuid] || 
   [Researcher, Admin].find{ |Klass| Klass.find_by(uuid: uuid)&.to_json(methods: [:is_admin, :is_researcher]) }
```

backend/app/services/user_info.rb
chrisbendel: RE: underscores, i didn't even notice this, these were all copy pasted from the dev/users_controller so those must have been in here for a while 😨 my JS brain just glanced over it, good catch!

These two methods are for the dev users list, so we aren't actually loading all the researchers anywhere in reality, plus there will be a limited list locally.

nathanstitt (approved):

PR #340, opened 2024-01-23
Only show active launched studies

chrisbendel:


frontend/src/screens/researcher/studies/create/edit-study.tsx
```diff
@@ -122,9 +123,12 @@ const FormContent: FC<{
     const [maxStep, setMaxStep] = useLocalstorageState<StudyStep>(`study-max-progress-${id}`, 0)
     const { isValid, isDirty } = useFormState()
 
-    if (!isDraft(study) && !isNew) {
-        return <Navigate to={`/study/overview/${id}`} />
-    }
+    useEffect(() => {
+        if (!isDraft(study) && !isNew) {
```
chrisbendel, line 127: There was some really weird inconsistent issue plaguing the playwright tests where the study would save, but it would redirect to the study overview page rather than the study creation flow... i believe this should finally fix that

frontend/src/screens/learner/details.tsx
```diff
@@ -193,7 +196,9 @@ export const StudyDetails: React.FC<{ studies: ParticipantStudy[] }> = ({ studie
         }
     }, [study])
 
-    if (!study) return null
+    if (isLoading) return null
+
+    if (!study) return <Navigate to="/studies" />
```
chrisbendel, line 201: Changes in this file are due to navigating directly into the details page, the studies weren't loaded yet and it wasn't rendering the component in time, i like the idea of fetching the individual study too on opening in case something with the study changes, but i know there is also an argument for filtering it out through the main studies list 🤔

backend/app/controllers/api/v1/participant/studies_controller.rb
```diff
@@ -5,7 +5,7 @@ class Api::V1::Participant::StudiesController < Api::V1::BaseController
   before_action :set_study, only: [:launch, :land, :stats]
 
   def index
-    launched_studies = current_user.launched_studies.includes(:stages, study: [:researchers])
+    launched_studies = current_user.active_launched_studies.includes(:stages, study: [:researchers])
```
nathanstitt, line 9: I don't think this is correct.

The `launched_studies` previously included all studies the user had ever launched, and this change will only include them if they're active, so should probably be renamed to `available_launched_studies`

This is then added to the `unlaunched_studies` below which is also poorly named (should probably be `available_studies`) which is a list of all the studies that are available. 

Since the 2nd variable already includes all available studies, the "available_launched_studies" will already be included in it.

I think what we intend to happen here is to include studies that are not available but have been *completed*, right?  In that case the 1st query should be rewritten to be: `completed_studies = current_user.launched_studies.includes(:stages, study: [:researchers]).complete` 

Its likely I've misread something though, and I'm not sure ☝️ is very clear 😁

nathanstitt: oh and I was probably the one who named all this, so that's on me 😭

iccgoncalves:
@nathanstitt @chrisbendel upon review of the above logic, let me share ticket objectives for clear path forward.

**Overview of the Task:**
Adjust logic to ensure that unavailable studies that have not been completed by the user do not show on their studies dashboard

**Requirements:** 

> Current Behavior
> Unavailable studies that have been launched but not completed are showing on the user's dashboard

> Expected Behavior
> We must show studies that have been completed (irregardless as to whether these are available or not)
> We must not show studies that are no longer available IF not completed

backend/app/controllers/api/v1/participant/studies_controller.rb
```diff
@@ -20,10 +15,15 @@ def index
   end
 
   def show
-    model =
-      current_user.launched_studies.where(study_id: params[:id]).first ||
-      Study.available_to_participants.find(params[:id])
-    raise ActiveRecord::RecordNotFound if model.is_hidden?
+    model = participant_studies.find do |study|
```
chrisbendel, line 18: Wanted to keep this method in sync with what the list would return (ie this method should only ever be filtering from a list of whats shown to them on the dashboard 🤔 ) hence these changes

backend/spec/requests/api/v1/participant/studies_spec.rb
```diff
@@ -144,6 +145,14 @@
         api_get 'participant/studies'
         expect(response_hash[:data]).not_to include(a_hash_including(id: study1.id))
       end
+
+      it 'excludes completed launched studies' do
+        # End both sessions
+        study2.end
+        study2.end
```
nathanstitt, line 152: this confused me for a bit, might be better to do:
`study2.stages.each{|stage| stage.update!(status: 'completed') }`

nathanstitt (approved):

backend/spec/requests/api/v1/participant/studies_spec.rb
chrisbendel: I would prefer to keep using `study.end` as it will uncover any issues if we change the `end` method down the line - not sure how strongly you feel about it either way though

the end method:
```
  def end
    stages.where.not(status: 'completed').first&.update!(status: 'completed')
  end
```

PR #343, opened 2024-02-07
Eng 38 interval bug

chrisbendel:


backend/spec/requests/api/v1/researcher/studies_spec.rb
chrisbendel, line 1: Updated these tests to support multiple sessions for more robustness

frontend/src/models/user.ts
```diff
@@ -19,7 +19,7 @@ export class User {
     id?: string = ''
     isAdmin: boolean = false
     isResearcher: boolean = false
-    name: string = ''
+    fullName: string = ''
```
chrisbendel, line 22: missed from an old PR - dev/index screen was missing the names

nathanstitt (approved):

PR #353, opened 2024-02-22
Admin manage study paths

chrisbendel:


frontend/src/screens/admin/manage-learning-paths.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+                <Group justify="flex-end" mt="md">
+                    <Button type="submit" disabled={!form.isValid()}>
+                        Create Learning Path
+                    </Button>
+                </Group>
+            </form>
+        </>
+    )
+}
```
nathanstitt, line 83: this seems like it's poorly typed and we're doing more work than needed.   Typically when I find myself in this situation it's better to let typescript do the inference.  this seems to work better for me with everything still being typed.

By moving the submit function inside the form.onSubmit, the values get typed due to form being typed.

```suggestion
const CreateLearningPath = () => {
    const form = useForm({
        initialValues: {
            label: '',
            description: '',
        } as LearningPath,
        validate: yupResolver(validationSchema),
    })

    const createLearningPath = useCreateLearningPath()

    const handleSubmit = form.onSubmit((values) => {
        createLearningPath.mutate({
            learningPath: values,
        }, {
            onSuccess: () => {
                form.reset()
            },
        })
    })

    return (
        <form onSubmit={handleSubmit}>
            <Stack>
                <TextInput
                    withAsterisk
                    label="Label"
                    {...form.getInputProps('label')}
                />

                <Textarea
                    withAsterisk
                    label="Description"
                    {...form.getInputProps('description')}
                />
            </Stack>

            <Group justify="flex-end" mt="md">
                <Button type="submit" disabled={!form.isValid()}>
                    Create Learning Path
                </Button>
            </Group>
        </form>
    )
}

```

frontend/src/screens/admin/manage-learning-paths.tsx
```diff
@@ -0,0 +1,132 @@
...
+const EditLearningPath: FC<{learningPath: LearningPath}> = ({ learningPath }) => {
+    const form = useForm<LearningPath>({
+        initialValues: {
+            label: learningPath.label,
+            description: learningPath.description,
+        },
+        validate: yupResolver(validationSchema),
+    })
+
+    type FormValues = typeof form.values;
```
nathanstitt, line 94: same as above, no need to setup a type here

nathanstitt (approved):
looking good, take look at the typing stuff and see what you think

PR #357, opened 2024-03-21
Open badge integration

chrisbendel:


backend/.gitignore
```diff
@@ -3,7 +3,6 @@
 
 # Ignore master key for decrypting credentials and more.
 /config/master.key
-/config/secrets.yml
```
chrisbendel, line 6: accidentally gitignored in the parent branch

frontend/src/screens/learner/studies.ts
```diff
@@ -52,13 +51,10 @@ export const useParticipantStudies = () => {
         allHighlightedStudies.length == FEATURED_COUNT ? [] : randomlyHighlighted
     )
 
-    const nonHighlightedStudies = studies.filter(s => !highlightedStudies.includes(s))
```
chrisbendel, line 55: With study paths, highlighted studies can be shown in both places now, so we dont need to filter them out from the rest of the studies below the highlighted section

backend/config/secrets.yml
```diff
@@ -13,7 +12,6 @@ development:
   enclave_api_key: ec_dev_1bG346BDBGEsG5Dck8H6sAiRBJ7Vt52Vwcno
   homepage_url: <%= ENV.fetch('HOMEPAGE_URL', 'http://localhost:4000') %>
   frontend_url: http://localhost:4000
-  admin_uuids: <%= ENV.fetch('ADMIN_UUIDS', '00000000-0000-0000-0000-000000000000') %>
```
chrisbendel, line 16: was unused

backend/db/migrate/20240229135718_migrate_study_learning_paths.rb
```diff
@@ -1,43 +1,123 @@
 class MigrateStudyLearningPaths < ActiveRecord::Migration[6.1]
```
chrisbendel, line 1: All data here per [this doc](https://docs.google.com/document/d/1nzGtI-DA2mDzNRuns2ZlSocvC7uH21qMmn5FT6CX5Os/edit)

backend/spec/factories/learning_path.rb
```diff
@@ -4,5 +4,6 @@
   factory :learning_path do
     label { Faker::Hobby.activity }
     description { Faker::Hobby.activity }
+    badge_id { 'SAJSINa7DGDaC4D' }
```
chrisbendel, line 7: a test badge id we have in OBF

docker-compose.yml
```diff
@@ -59,16 +60,17 @@ services:
...
-    image: "swaggerapi/petstore3"
-    networks:
-      - kinetic
-    environment:
-      - SWAGGER_HOST=http://0.0.0.0:4008
-      - SWAGGER_BASE_PATH=/api/v1
-      - SWAGGER_URL=http://0.0.0.0:4006
-    ports:
-      - "4008:8080"
+#  TODO: Investigate if needed
```
chrisbendel, line 63: Wasn't sure if anything was necessarily using this, tried looking through the code but couldn't find anything. It has a slow startup/shutdown time with docker compose and we get some warnings about something with the container image 🤔

frontend/src/components/segment-bar.tsx
```diff
@@ -150,7 +150,7 @@ export const SegmentedBar: FCWC<SegmentedBarProps> = ({ className, children, com
             >
                 <span
                     data-testid="progress-indicator"
-                    data-percentage-complete={completedPercentage}
+                    data-percentage-complete={completedPercentage || ''}
```
chrisbendel, line 153: Fixes a console warning/error when no rewards are setup

backend/app/services/qualtrics_api.rb
```diff
@@ -6,7 +6,9 @@
 class QualtricsApi
 
   def initialize
-    @http = HTTPX.plugin(:compression).with(
+    puts('qualtrics api key')
+    puts(Rails.application.secrets.qualtrics_api_key)
```
nathanstitt: are these debugging?

backend/db/migrate/20240229135718_migrate_study_learning_paths.rb
```diff
@@ -1,43 +1,123 @@
...
   def up
```
nathanstitt, line 2: this is an existing migration.  modifying it will have no effect if it's already been ran, like on staging & production

backend/db/structure.sql
```diff
@@ -152,8 +152,8 @@ ALTER SEQUENCE public.active_storage_variant_records_id_seq OWNED BY public.acti
 CREATE TABLE public.admins (
     id bigint NOT NULL,
     user_id uuid NOT NULL,
-    created_at timestamp with time zone NOT NULL,
-    updated_at timestamp with time zone NOT NULL
+    created_at timestamp(6) with time zone NOT NULL,
+    updated_at timestamp(6) with time zone NOT NULL
```
nathanstitt, line 156: we had this occur before,  as I recall it's caused by rails defaults for timestamp.  maybe it's a side effect of the rails 7 migration?

docker-compose.yml
```diff
@@ -59,16 +60,17 @@ services:
...
+#  petstore:
+#    image: "swaggerapi/petstore3"
+#    networks:
+#      - kinetic
+#    environment:
+#      - SWAGGER_HOST=http://0.0.0.0:4008
+#      - SWAGGER_BASE_PATH=/api/v1
+#      - SWAGGER_URL=http://0.0.0.0:4006
+#    ports:
+#      - "4008:8080"
```
nathanstitt, line 73: `petstore` is just a graphical browser for the open api definitions.  it's sometimes nice to use to debug things, but doesn't need to run all the time.  You could move the definition to a separate compose file and add the directions to the README so it could be started when needed: https://stackoverflow.com/questions/45680958/can-you-define-optional-docker-compose-services

frontend/src/screens/learner/studies.ts
```diff
@@ -52,13 +52,13 @@ export const useParticipantStudies = () => {
...
+    // const nonHighlightedStudies = studies.filter(s => !highlightedStudies.includes(s))
```
nathanstitt: should remove line if not needed

frontend/src/screens/learner/studies.ts
```diff
@@ -52,13 +52,13 @@ export const useParticipantStudies = () => {
...
 
     const demographicSurvey = studies.find(s => s.isDemographicSurvey) || null
 
     return {
         allStudies: studies,
-        nonHighlightedStudies,
+        // nonHighlightedStudies,
```
nathanstitt: same as above

nathanstitt (approved):
looks good but you'll need to make a new migration vs modifying the existing one

PR #360, opened 2024-03-28
use find_each instead of each

chrisbendel:


backend/app/services/learner_activity_report.rb
```diff
@@ -33,7 +33,7 @@ def build_headers(csv)
   end
 
   def get_users(launches)
-    launches.each do |launch|
+    launches.find_each do |launch|
```
nathanstitt, line 36: since all we're reading is the `user_id` we could use `pluck`

```suggestion
    @user_uuids = launches.clone.pluck('user_id')
```
The `clone` is needed because launches is a query that's also used below, so we create a clone before executing it.

nathanstitt (approved):

PR #364, opened 2024-06-06
Cleanup ENV

chrisbendel:
The new interns are both on windows machines, they were running into some issues with the `frontend/bin/env` file. I tried to investigate if we could potentially just remove that file and rely on env variables. it seems the main purpose of `bin/env` was to pass along the `VITE_<var>` env variables for staging/prod.

The GA_UA tags should be irrelevant now as that was the former version of GA that is no longer used.
The GTM tag is shared between prod and staging (checked with Tom on that), so simplified that down to one config
I also set the `VITE_GTAG_ID` and `VITE_SENTRY_DSN` on staging/prod preemptively for whenever this gets merged


nathanstitt (approved):
as i recall some of the bin/env was needed to setup things on heroku, but as long as this all builds on staging i think this is fine

PR #367, opened 2024-06-13
toast-notification created

divitcr7:
Please delete this PR


frontend/src/screens/study-landing.tsx
```diff
@@ -1,137 +1,236 @@
...
-        >
-            <Card p={{ lg: 'xl' }} m={{ lg: 'xl' }} shadow='md' radius='md'>
-                {study.learningPath?.completed ?
-                    <CompletedLearningPath learningPath={study.learningPath} /> :
-                    <LearningPathProgress learningPath={study.learningPath} studies={learningPathStudies} />
-                }
-            </Card>
-        </Page>
-    )
+        <MantineProvider>
```
chrisbendel, line 98: We already have a <MantineProvider> at the root [here](https://github.com/openstax/kinetic/blob/main/frontend/src/main.tsx#L30) additionally in this file the <Notifications> container is also created. You shouldn't really have to write any new styles or JSX, as we can just use the [notifications system](https://mantine.dev/x/notifications/)  [example here](https://github.com/openstax/kinetic/blob/main/frontend/src/components/notifications/researcher-notification.tsx)

let me know if this makes sense! i think we can probably simplify this PR quite a bit with those changes :)

divitcr7:
Thank you, I'll fix it

chrisbendel:
@divitcr7 is this PR stale / unneeded now? if so - we should close this PR and delete the branch to clean things up

PR #368, opened 2024-06-14
Added the quick link section

Coder-Srinivas:
I have added the quick links section as per the ticket. Let me know if any changes are required. 

https://github.com/openstax/kinetic/assets/59244289/2114c0d8-9dd2-489c-8cce-e0c67b74e65e


frontend/src/screens/learner.tsx
```diff
@@ -172,7 +172,6 @@ export const DesktopStudyCards: FC<{studies: ParticipantStudy[]}> = ({ studies }
             <Swiper
                 slidesPerView={3}
                 simulateTouch={true}
-                freeMode={true}
```
chrisbendel, line 175: would you be able to explain why you removed this?

frontend/src/screens/learner/card.tsx
```diff
@@ -15,17 +15,17 @@ interface StudyCardProps {
 }
 
 const Card = styled(Box)({
-    minWidth: 400,
-    maxWidth: 400,
+    minWidth: 272,
```
chrisbendel, line 18: you should double check with @dkhowala about this, this will affect the card size throughout the entire application and we may not want this to be global, but if we want to change the general card size anyway this should be fine

frontend/src/screens/learner.tsx
```diff
@@ -201,45 +199,99 @@ export const DesktopStudyCards: FC<{studies: ParticipantStudy[]}> = ({ studies }
...
+        [key: string]: HTMLElement | null;
+    };
+
+    const learningPathRefs = useRef<LearningPathRefs>({})
+
+    const scrollToLearningPath = (learningPath: string) => {
+        learningPathRefs.current[learningPath]?.scrollIntoView({ behavior: 'smooth', block: 'start' })
+    }
+
+    const [hoveredLearningPath, setHoveredLearningPath] = useState<string | null>(null)
```
chrisbendel, line 230: is this logic/css update required in the designs? i wonder if we could do this with css instead of useRef? in general i try to avoid useref as much as possible cause its one more thing to manage

frontend/src/screens/learner.tsx
Coder-Srinivas: The slides were overlapping, so I wanted to see if any of the properties in the swiper component are causing the issue, so  I removed everything and kept checking if it would solve the issue. I found that the overlapping issue would be solved if the width of the cards were reduced, freeMode can be added back. I overlooked that.

frontend/src/screens/learner/card.tsx
Coder-Srinivas: I will check with devika.

Coder-Srinivas:
I have confirmed with Devika, she said that `width: Fixed (264px)px height: Fixed (350px)px;` is for all the cards. 

Multi session cards will have `width: Fixed (272px)px; height: Fixed (358px)px;`

frontend/src/screens/learner.tsx
```diff
@@ -201,45 +200,93 @@ export const DesktopStudyCards: FC<{studies: ParticipantStudy[]}> = ({ studies }
...
-                            <DesktopStudyCards studies={studies} />
-                        }
-                    </Stack>
-                )
-            })}
-        </Stack>
+                    )
+                })}
+            </Box>
+            <Box style={{ width: '75%', gap: 'lg' }} data-testid='studies-listing'>
```
chrisbendel, line 261: @Coder-Srinivas  When possible we should try to use <Stack> or <Group>, or in really custom scenarios <Flex> instead of box as it will have a lot of the flex properties available as props (ie `gap` in this instance) and they [can take shorthand CSS properties](https://mantine.dev/styles/style-props/)

Coder-Srinivas: Got it.

chrisbendel (approved):

divitcr7 (approved):
Works as intended

PR #371, opened 2024-06-19
notification toast

divitcr7:


chrisbendel:
Hey @divitcr7 i don't think we were supposed to remove the existing landing page. Instead I believe we wanted to add a new component for studies that didn't have a learning path? You should not be removing anything just adding something new for a new condition. Let me know if that makes sense

PR #372, opened 2024-06-19
all changes in the 2nd ticket

divitcr7:


chrisbendel:
Also these look like the same exact changes that were made in #371

PR #374, opened 2024-06-28
RENG-105

divitcr7:
Ticket link : https://openstax.atlassian.net/browse/RENG-105


chrisbendel:
@divitcr7 it looks like a lot of changes were going from single quotes `'` to double quotes `"` which makes the pull request harder to read 

We have a rule in [`eslintrc.js` on line 68](https://github.com/openstax/kinetic/blob/main/frontend/.eslintrc.js#L68) which enforces single quotes, if you can set up your editor to use the eslint rules the repository has - then I can get around to reviewing this, thanks!

PR #376, opened 2024-06-29
all checks + changes

divitcr7:
Ticket -> https://github.com/openstax/kinetic/pull/374

### DIFF
--- a/frontend/src/components/footer.tsx
+++ b/frontend/src/components/footer.tsx
@@ -1,8 +1,7 @@
 import { React } from '@common'
 import { HelpLink, ResourceLinks } from './resource-links'
-import { colors } from '@theme'
 import { useIsMobileDevice } from '@lib';
-import { Divider,Anchor, Box, Container, Flex, Group, Image, Stack, Text, Title } from '@mantine/core';
+import { Divider, Anchor, Box, Container, Flex, Group, Image, Stack, Text, Title } from '@mantine/core';
 import { IconBrandFacebookFilled, IconBrandInstagram, IconBrandTwitterFilled } from '@tabler/icons-react';
 
 // @ts-ignore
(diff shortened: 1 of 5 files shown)


divitcr7:
https://openstax.atlassian.net/browse/RENG-105

frontend/src/components/footer.tsx
```diff
@@ -16,31 +16,90 @@ import OpenStaxURL from '../images/openstax-logo.png'
...
-                <Stack gap='md' mb='md'>
+    
+                    <Divider orientation="vertical" style={{ height: '200px', width: '1px', borderWidth: '1px' }} />
+    
+                    <Anchor target="_blank" href="https://ies.ed.gov/">
+                        <Stack
+                            align="center"
+                            gap="sm"
+                            style={{
+                                width: '462.66px',
```
chrisbendel, line 62: I would avoid using hardcoded widths (unless absolutely necessary) we want these to be somewhat responsive on mobile just in case they don't resize properly

Also keep in mind these `[mantine](https://mantine.dev/)` components have a lot of props for common css properties (ie above you replaced bg with a style={{background}} and here we have a gap `sm` property and you would be overwriting it with gap: 16px, theres also padding with the `p` prop

frontend/src/screens/learner.tsx
```diff
@@ -126,17 +133,29 @@ export const StudiesContainer = () => {
                     <SearchBar search={search} setSearch={setSearch} />
                 </Flex>
 
+                {/* Divider */}
```
chrisbendel, line 136: https://mantine.dev/core/divider/

do you want to just use this instead? consistent with the other uses in here

frontend/src/screens/learner.tsx
```diff
@@ -145,81 +164,95 @@ export const MobileStudyCards: FC<{studies: ParticipantStudy[]}> = ({ studies })
...
+                navigation={true}
             >
-                {studies.map((study) => (
+                {studies.sort((a, b) => (a.completedAt === b.completedAt) ? 0 : a.completedAt ? 1 : -1).map((study) => (
                     <SwiperSlide key={study.id} className="pb-1">
                         <StudyCard study={study} />
                     </SwiperSlide>
                 ))}
             </Swiper>
+            <style>
```
chrisbendel, line 181: we probably don't want inline CSS here - could we extract this out to a .css file? we may already have one - worth checking

frontend/src/screens/learner.tsx
```diff
@@ -145,81 +164,95 @@ export const MobileStudyCards: FC<{studies: ParticipantStudy[]}> = ({ studies })
...
-    )
-}
+    );
+};
 
-export const DesktopStudyCards: FC<{studies: ParticipantStudy[]}> = ({ studies }) => {
+export const DesktopStudyCards: FC<{ studies: ParticipantStudy[] }> = ({ studies }) => {
     return (
-        <Box>
+        <Box style={{ position: 'relative', '&:hover .swiper-button-next, &:hover .swiper-button-prev': { opacity: 1 } }}>
```
chrisbendel, line 200: same thing as CSS comment above ^

frontend/src/screens/learner.tsx
```diff
@@ -145,81 +164,95 @@ export const MobileStudyCards: FC<{studies: ParticipantStudy[]}> = ({ studies })
...
+                modules={[FreeMode, Navigation]}
             >
-                {studies.map(study => (
+                {studies.sort((a, b) => (a.completedAt === b.completedAt) ? 0 : a.completedAt ? 1 : -1).map((study) => (
                     <SwiperSlide style={{ padding: '1rem' }} key={study.id}>
                         <StudyCard study={study} />
                     </SwiperSlide>
                 ))}
             </Swiper>
+            <style>
```
chrisbendel, line 220: css

PR #377, opened 2024-07-01
Remove PII

chrisbendel:


RoyEJohnson (approved):
Were any calls to `find_user_email` removed previously?

chrisbendel:
@RoyEJohnson yeah we had removed an email column and just never ended up removing that email method, so its just cleanup here

PR #380, opened 2024-07-05
Added a feature to sort studies based on duration

Coder-Srinivas:
Added the feature based on the following [ticket](https://openstax.atlassian.net/browse/RENG-104?atlOrigin=eyJpIjoiNGNjZWU3ZmFjMzk4NDc4ZjkzMTdiZGQyNzRjYmYxZGEiLCJwIjoiaiJ9)

[Design Spec](https://www.figma.com/design/MVxS1h1qovYyDqbWf3DTyF/Version-3.0-Apr24?node-id=1458-2632&t=hvSypsZ5u0YtLk6G-0)

The learner can now sort based on the duration of the study. (5 min or 15min or 25min).

### DIFF
--- a/frontend/src/components/icon.tsx
+++ b/frontend/src/components/icon.tsx
@@ -50,7 +50,6 @@ export const ICONS = {
     clockOutline,
     plusCircleOutline,
     plus,
-    minus,
     close,
     search,
     trash,
--- a/frontend/src/components/icon.tsx
+++ b/frontend/src/components/icon.tsx
@@ -50,6 +50,7 @@ export const ICONS = {
     clockOutline,
     plusCircleOutline,
     plus,
+    minus,
     close,
     search,
     trash,
--- a/frontend/src/screens/learner.tsx
+++ b/frontend/src/screens/learner.tsx
@@ -1,6 +1,6 @@
 import { React, styled } from '@common'
 import { ParticipantStudy } from '@api'
-import { Footer, TopNavBar, Icon } from '@components'
+import { Footer, TopNavBar } from '@components'
 import { useEnvironment, useIsMobileDevice } from '@lib'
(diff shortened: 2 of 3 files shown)


frontend/src/components/icon.tsx
```diff
@@ -50,6 +50,7 @@ export const ICONS = {
     clockOutline,
     plusCircleOutline,
     plus,
+    minus,
```
chrisbendel, line 53: See icon note below

frontend/src/screens/learner.tsx
```diff
@@ -114,15 +114,67 @@ export const SearchResults: FC<{search: string, filteredStudies: ParticipantStud
...
+    return (
+        <Flex justify='center' align='center' gap='.5rem' 
+            pt='.25rem' pb='.25rem' pl='.625rem' pr='.625rem'
+            bg={ active? colors.blue: colors.white }
+            style={{ border: `1px solid ${colors.blue}`, borderRadius: '50rem', transition: 'all .1s ease-in' }} 
+            onClick={() => {
+                handleDurationChange(durationText)
+            }}>
+            <Text size='sm' c={ active? colors.white: colors.blue }>~{String(durationText)} min</Text> 
+            <Circle><Icon height={10} color={colors.blue} icon={active? 'minus': 'plus'}></Icon></Circle>
```
chrisbendel, line 161: For Icons, i think we want to try and move towards using `@tabler/icons-react`, trying ton consolidate and eventually remove uses of this `<Icon>` component here - let me know if that makes sense

frontend/src/components/icon.tsx
chrisbendel, line 53: Thnk we can remove this if nowhere else is using it

chrisbendel (approved):

PR #382, opened 2024-07-12
Added an option to add color to the learning path

Coder-Srinivas:
Added an option for the admin to add color for a learning path, based on the [ticket](https://openstax.atlassian.net/browse/RENG-117?atlOrigin=eyJpIjoiMzkxZWUyZmRmODUwNDMyMzlhNzIwNmQ4MjQ3YmRmNTUiLCJwIjoiaiJ9)

- Used combobox as the input rather than select as each option can be modified individually.
- Study cards PR needs to be updated once this gets merged, as learning path color is given by the server.

### DIFF
--- a/backend/db/migrate/20240222135611_create_study_learning_paths.rb
+++ b/backend/db/migrate/20240222135611_create_study_learning_paths.rb
@@ -4,6 +4,7 @@ class CreateStudyLearningPaths < ActiveRecord::Migration[6.1]
       t.string :label, null: false
       t.string :description, null: false
       t.string :badge_id
+      t.string :color, null: false
       t.text :level_1_metadata, array: true, default: []
       t.text :level_2_metadata, array: true, default: []
       t.timestamps
--- a/backend/db/migrate/20240222135611_create_study_learning_paths.rb
+++ b/backend/db/migrate/20240222135611_create_study_learning_paths.rb
@@ -4,7 +4,6 @@ class CreateStudyLearningPaths < ActiveRecord::Migration[6.1]
       t.string :label, null: false
       t.string :description, null: false
       t.string :badge_id
-      t.string :color, null: false
       t.text :level_1_metadata, array: true, default: []
       t.text :level_2_metadata, array: true, default: []
       t.timestamps
--- a/backend/db/structure.sql
+++ b/backend/db/structure.sql
@@ -432,6 +432,7 @@ CREATE TABLE public.learning_paths (
     label character varying NOT NULL,
     description character varying NOT NULL,
     badge_id character varying,
+    color character varying NOT NULL,
     level_1_metadata text[] DEFAULT '{}'::text[],
     level_2_metadata text[] DEFAULT '{}'::text[],
     created_at timestamp(6) with time zone NOT NULL,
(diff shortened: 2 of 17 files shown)


backend/db/migrate/20240222135611_create_study_learning_paths.rb
```diff
@@ -4,6 +4,7 @@ def change
       t.string :label, null: false
       t.string :description, null: false
       t.string :badge_id
+      t.string :color, null: false
```
chrisbendel, line 7: @divitcr7 rather than adding this to an old file, we will need to add a new migration to add the new column. check this documentation on [creating a new migration here](https://guides.rubyonrails.org/active_record_migrations.html#generating-migrations) and let me know if you have any questions. It should be pretty simple for adding just one column :) 

You will have to remove these changes from this file and the `populate_learning_path_data.rb` file as well

backend/db/structure.sql
```diff
@@ -432,6 +432,7 @@ CREATE TABLE public.learning_paths (
     label character varying NOT NULL,
     description character varying NOT NULL,
     badge_id character varying,
+    color character varying NOT NULL,
```
chrisbendel, line 435: this will get auto generated once you run the new migration file, so remove this

test-results/.last-run.json
chrisbendel, line 1: should remove this file or add it to `.gitignore` if this will keep resurfacing

frontend/src/screens/admin/manage-learning-paths.tsx
```diff
@@ -309,6 +320,41 @@ const LearningPathForm: FC<{
                     label='Order'
                     {...form.getInputProps('order')}
                 />
+
```
chrisbendel, line 323: One question for @iccgoncalves is will we allow for users to select custom colors as well? or only from this predefined list.

Regardless, mantine has a nifty little color input component that you can disallow input on and just [set swatches of colors as the options](https://mantine.dev/core/color-input/#with-swatches)

We can just use that `disallowInput` prop whether or not we want them to have the ability to store custom colors. let me know what you think of this vs the combobox

chrisbendel:
Looking good - just have to hit a couple of the comments I left! 😄

frontend/src/screens/admin/manage-learning-paths.tsx
iccgoncalves: Thank you for this question @chrisbendel ;) 
We do not want to allow for custom color selection - the provided list of colors is associated with the badges color, so we want to keep it within that list. Thanks so much!

Coder-Srinivas: @chrisbendel Should I continue with the combobox or use the color input component

test-results/.last-run.json
Coder-Srinivas: Okay

frontend/src/screens/admin/manage-learning-paths.tsx
chrisbendel: @Coder-Srinivas i kinda like the look of the color picker - it looks a little simpler visually. would you be willing to try it out and see how it goes?

chrisbendel: I think the implementation of the component will be a bit simpler too, compared to the combobox

Coder-Srinivas:
@chrisbendel can you please review this PR, as this needs to be merged first for the study card redesign to be merged, as the borders for the mutli session cards are based on the color of the learning path

chrisbendel (approved):
Looks good - feel free to merge!

PR #384, opened 2024-07-15
Ticket-88

divitcr7:
added notification banner after completion of study

### DIFF
--- a/frontend/src/screens/study-landing.tsx
+++ b/frontend/src/screens/study-landing.tsx
@@ -1,5 +1,5 @@
 import { Navigate, NavLink, useLoaderData } from 'react-router-dom'
-import { React, useEffect, useState } from '@common'
+import { React, useEffect } from '@common'
 import { colors } from '@theme'
 import { LearningPath, ParticipantStudy } from '@api'
 import { Page } from '@components'
@@ -28,11 +28,8 @@ export default function StudyLanding() {
     const env = useEnvironment()
     const study = useLoaderData() as ParticipantStudy
     const learningPathStudies = useLearningPathStudies(study?.learningPath)
-    const [notificationShown, setNotificationShown] = useState(false);
 
     const showEarnedPointsNotification = (points: number) => {
-        if (notificationShown) return;
-        
         notifications.show({
             title: `You just earned ${points} points!`,
             message: 'The longer the study, the more points you earn. Reach 200 points to unlock additional rewards.',
@@ -43,25 +40,21 @@ export default function StudyLanding() {
                 description: { fontSize: '12px' },
             }),
         });
-        
-        setNotificationShown(true);
     };
 
     useEffect(() => {
-        if (!notificationShown && study.totalPoints > 0) {
+        if (study.totalPoints > 0) {
             const timer = setTimeout(() => {
                 showEarnedPointsNotification(study.totalPoints);
             }, 100);
             
             return () => clearTimeout(timer);
         }
-    }, [notificationShown, study.totalPoints]);
-
+    }, [study.totalPoints]);
 
     if (!study || !study.learningPath) {
         return <Navigate to='/studies' />
     }
-
 
     return (
         <Page hideFooter
--- a/frontend/src/screens/study-landing.tsx
+++ b/frontend/src/screens/study-landing.tsx
@@ -1,5 +1,5 @@
 import { Navigate, NavLink, useLoaderData } from 'react-router-dom'
-import { React, useEffect } from '@common'
+import { React, useEffect, useState } from '@common'
 import { colors } from '@theme'
 import { LearningPath, ParticipantStudy } from '@api'
 import { Page } from '@components'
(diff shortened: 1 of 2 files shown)


chrisbendel:
@divitcr7 also could you please update the PR title to something more descriptive, thanks!

divitcr7:
This ticket details was changed initially after the task was done, 
https://openstax.atlassian.net/browse/RENG-88

as per the ticket we were first supposed to create a study banner, but as per the discussion only the notifications banner on the top right of the screen was supposed to be added , tht stated the no. of points

frontend/src/screens/study-landing.tsx
```diff
@@ -100,7 +137,7 @@ const CompletedLearningPath: FC<{learningPath: LearningPath}> = ({ learningPath
 
                     <Stack gap='0'>
                         <Title order={5} c='purple'>
-                            You’ve done awesome work so far!
+                        You’ve done awesome work so far!
```
chrisbendel, line 140: weird formatting here - why remove the indent?

frontend/src/screens/study-landing.tsx
```diff
@@ -44,6 +62,25 @@ export default function StudyLanding() {
...
+                        position: 'absolute',
+                        top: '1rem',
+                        right: '1rem',
+                        zIndex: 1000,
+                        maxWidth: '400px',
+                    }}
+                >
+                    <Title order={6} mb={5}>You just earned {study.totalPoints} points!</Title>
+                    <Text size="xs" lineClamp={2} c={colors.gray70}>
+                    The longer the study, the more points you earn. Reach 200 points to unlock additional rewards.
```
chrisbendel, line 80: formatting here too, why was the indent removed 🤔

frontend/src/screens/study-landing.tsx
```diff
@@ -44,6 +62,25 @@ export default function StudyLanding() {
                     <LearningPathProgress learningPath={study.learningPath} studies={learningPathStudies} />
                 }
             </Card>
+            {showNotification && (
```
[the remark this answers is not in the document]
divitcr7: used mantine notifications.show() as it was used in, the code base

frontend/src/screens/study-landing.tsx
```diff
@@ -18,20 +18,44 @@ import {
...
 import { useLearningPathStudies } from './learner/studies';
 import { CompactStudyCard } from '../components/study/compact-study-card';
+import { notifications } from '@mantine/notifications';
 
 export default function StudyLanding() {
     const env = useEnvironment()
-
     const study = useLoaderData() as ParticipantStudy
     const learningPathStudies = useLearningPathStudies(study?.learningPath)
+    const notificationShown = useRef(false)
```
chrisbendel, line 31: Is there a simpler way to do this without using `useRef`? I think the user will just land on this page once, you can probably change the useEffect below to just have an empty dependency array so that it only shows once? Let me know if that works, should simplify this a bit :) @divitcr7

frontend/src/screens/study-landing.tsx
```diff
@@ -18,20 +18,59 @@ import {
     Text,
     Title,
 } from '@mantine/core';
+import { IconCheck } from '@tabler/icons-react'
 import Markdown from 'react-markdown'
 import { useLearningPathStudies } from './learner/studies';
 import { CompactStudyCard } from '../components/study/compact-study-card';
+import { notifications } from '@mantine/notifications';
+
+let notificationShownThisSession = false;
```
chrisbendel, line 27: I dont think we should be doing it this way - we don't want to be modifying a variable outside of the component state.

frontend/src/screens/study-landing.tsx
```diff
@@ -18,20 +18,59 @@ import {
...
+    useEffect(() => {
+        if (study.totalPoints > 0) {
+            // Delay to ensure it runs after any potential double renders
+            const timer = setTimeout(() => {
+                showEarnedPointsNotification(study.totalPoints);
+            }, 100);
+            
+            return () => clearTimeout(timer);
+        }
+    }, [study.totalPoints]);
```
chrisbendel, line 60: what happens if you remove `study.totalPoints` from the dependency array? ideally this should just run once when the component renders initially right?

frontend/src/screens/study-landing.tsx
```diff
@@ -18,20 +18,59 @@ import {
...
+
+    // Reset the flag when the component unmounts
+    useEffect(() => {
+        return () => {
+            notificationShownThisSession = false;
```
chrisbendel, line 65: remove this RE: above comment

frontend/src/screens/study-landing.tsx
divitcr7: yes but, when i did that for some weird reason, the notifcations got displayed twice, one above another like this: 
image

chrisbendel: this could be just a side effect of [react rendering components twice in dev mode](https://chanduthedev.medium.com/why-react-app-rendering-twice-e31339ed9461#:~:text=StrictMode%20in%20the%20application%2C%20react,environment%2C%20it%20only%20renders%20once.&text=If%20you%20want%20to%20avoid,the%20development%20environment%2C%20remove%20React.)

chrisbendel:
@divitcr7 did you confirm if it was just the react strictmode/local dev server that was causing the double render? i think we should remove the `useState` for notificationShown if possible... should be unneeded if thats what it is? cc @Coder-Srinivas - curious what you think

nathanstitt:
i’ll jump in here since i think i know the cause, it’s likely because react
triggers hooks twice in dev mode.  search for “double useEffect” should
pull up docs

chrisbendel (approved):

PR #387, opened 2024-07-29
Fixed the obf email issue

Coder-Srinivas:
Added the authorization token to the http request and changed it from form to json as per the API documentation. @chrisbendel you were right, the email_address is handled differently in production and development. I have also modified the development userInfo object to match the production one


chrisbendel (approved):
great find!!

PR #391, opened 2024-08-05
Added color to the images based on the learning path

Coder-Srinivas:
- The color of the study images is now based on the color of the learning path.
- Added a placeholder image instead of using the AWS bucket.
- Removed the eye dropper from color input
- The secondary color of the images is replaced with the color of the learning path


frontend/src/components/study-card-images/study-image.tsx
```diff
@@ -0,0 +1,46 @@
+import { React, useState, useEffect } from '@common'
+import { ParticipantStudy } from '@api'
+import { getImageUrl, getAltText } from './card-images'
+import { colors } from '@theme';
+
+const SVGManipulator: React.FC<{src: string, fillColor: string, altText: string}> = ({ src, fillColor, altText }) => {
```
chrisbendel, line 6: I am not an expert with SVGs so im curious about this approach, it looks like we are modifying the `fill` property in the SVG and then replacing the image?

would this be possible/simpler with CSS? something like this? not sure if it work, but would simplify this a bit: https://stackoverflow.com/questions/9529300/can-i-change-the-fill-color-of-an-svg-path-with-css

frontend/src/components/study-card-images/study-image.tsx
```diff
@@ -0,0 +1,46 @@
...
+  
+    return <img src={imgSrc} alt={altText} className='study-card-image' />;
+};
+
+
+export const StudyCardImage: React.FC<{study: ParticipantStudy}> = ({ study }) => {
+
+    return (
+        <SVGManipulator 
+            src={getImageUrl(study.imageId)}
```
chrisbendel, line 41: More a question of where getImageUrl is defined - do we still want to reference the kinetic s3 bucket or are all the images officially local now with your placeholder addition here?

frontend/src/components/study-card-images/study-image.tsx
Coder-Srinivas: I tried that approach initially and found that since we are using an img tag and using the svg as a src, it would not work. For the above approach to work, the svg file should be rendered directly instead of using an img tag, since the images are in the public folder, I could not import them, to use them directly.

Coder-Srinivas: Another problem I noticed with that approach is that on setting the fill values on the svg container, it automatically applies to all the paths, unless a fill property is already present. So we have to manually edit all the paths which don't have a fill value and set it to none, and remove the fill value for the paths which have the secondary color. This way by setting the fill value once, it is applied to all the paths which don't have a fill value (secondary color). The only problem is, the svg files have to be edited manually and should be placed inside the src folder

frontend/src/components/study-card-images/study-image.tsx
Coder-Srinivas: getImageUrl is defined in the card-images.tsx file which is inside the study-cards-images component folder. The kinetic s3 bucket is no longer needed.

Coder-Srinivas:
I also noticed that, I would fail if the image is a png. There a couple of png images in the public folder.

chrisbendel (approved):

chrisbendel:
Great job! Feel free to merge this @Coder-Srinivas

PR #397, opened 2024-12-19
Modifed Admin Reports

Coder-Srinivas:
Removed the column stage points and added the columns study path title and completed study path, to the reports csv file

Also removed the feature of downloading reports and added the feature to email the reports to the admin, sent a notification when the email is sent successfully.

### DIFF
--- a/frontend/src/screens/admin/admin-reports.tsx
+++ b/frontend/src/screens/admin/admin-reports.tsx
@@ -1,6 +1,7 @@
 import { Main } from './grid';
 import { React, useState } from '@common';
 import { Button, Center, Group, NumberInput, Stack, Title } from '@mantine/core';
+import { useCurrentUser } from '@lib';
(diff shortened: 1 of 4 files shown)


frontend/src/screens/admin/admin-reports.tsx
```diff
@@ -25,8 +30,41 @@ const LearnerActivity = () => {
                         value={monthsAgo}
                         onChange={setMonthsAgo}
                     />
-                    <Button component='a' target='_blank' href={`${ENV.API_ADDRESS}/api/v1/admin/reports/learner-activity?months_ago=${monthsAgo}`}>
-                        Generate Report
+                    <Button onClick={async () => {
```
chrisbendel, line 33: This logic could probably be extracted out of the JSX and up in the top of the component to keep things cleaner 😄

PR #401, opened 2025-01-13
Added user notifications

Coder-Srinivas:
- Send a welcome mail to users when they finish their first study
- Send an email to users when a new study becomes available
- Send an email to users when a new learning path becomes available
- Send an email to inactive users, (when they are inactive for more than 60 days and an email hasn't been sent already)

### DIFF
--- a/backend/app/controllers/api/v1/participant/studies_controller.rb
+++ b/backend/app/controllers/api/v1/participant/studies_controller.rb
@@ -5,6 +5,9 @@ class Api::V1::Participant::StudiesController < Api::V1::BaseController
   before_action :set_study, only: [:launch, :land, :stats]
 
   def index
+    current_user_preference = UserPreferences.find_or_initialize_by(user_id: current_user.id)
+    current_user_preference.update!(last_visited_at: Time.current, nurturing_email_sent: false)
+    
     studies = participant_studies
 
     response_binding = Api::V1::Bindings::ParticipantStudies.new(
--- a/backend/app/models/launched_study.rb
+++ b/backend/app/models/launched_study.rb
@@ -30,6 +30,19 @@ class LaunchedStudy < ApplicationRecord
     update!(completed_at: Time.now)
     completed = LaunchedStudy.where(study_id:).complete.count
     Study.update(study_id, completed_count: completed)
+  
+    total_completed_for_user = LaunchedStudy.where(user_id:).count
+
+    user_info = UserInfo.for_uuid(user_id)
+  
+    recipient = Struct.new(:email_address, :full_name).new(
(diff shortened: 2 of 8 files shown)


backend/app/controllers/api/v1/participant/studies_controller.rb
```diff
@@ -5,6 +5,9 @@ class Api::V1::Participant::StudiesController < Api::V1::BaseController
   before_action :set_study, only: [:launch, :land, :stats]
 
   def index
+    current_user_preference = UserPreferences.find_or_initialize_by(user_id: current_user.id)
+    current_user_preference.update!(last_visited_at: Time.current, nurturing_email_sent: false)
```
chrisbendel, line 9: do we want to reset the `nurturing_email_sent` every time they load the studies page? just checking :)

backend/app/services/user_notifications.rb
```diff
@@ -37,7 +37,9 @@ def deliver_new_studies
       return unless users.any?
 
       users.each_value do |user|
-        UserMailer.with(user:, studies:).new_studies.deliver
+        studies.each do |study|
```
chrisbendel, line 40: with looping over each study, would we be emailing user A for every new study? ie would they get 5 emails if there were 5 new studies? is that what we want?

backend/config/initializers/schedule.rb
```diff
@@ -13,6 +13,73 @@ def populate_badges
...
+  return initiated_learning_paths.first, true if initiated_learning_paths.any?
+
+  initiated_learning_path_ids = LaunchedStudy
+                                  .where(user_id:)
+                                  .joins(:study)
+                                  .pluck('studies.learning_path_id')
+
+  selected_learning_path = LearningPath
+                             .where.not(id: initiated_learning_path_ids)
+                             .order('RANDOM()')
```
chrisbendel, line 33: just curious about this RANDOM() here 😄

backend/app/models/launched_study.rb
```diff
@@ -30,6 +30,18 @@ def completed!
...
     completed = LaunchedStudy.where(study_id:).complete.count
     Study.update(study_id, completed_count: completed)
+
+    total_completed_for_user = LaunchedStudy.where(user_id:).count
+
+    user_info = UserInfo.for_uuid(user_id)
+    recipient = Struct.new(:email_address, :full_name).new(
+      user_info['email_address'] || 'Admin-Uno@test.openstax.org',
+      user_info[:full_name]
+    )
```
chrisbendel, line 40: Technically this can be moved below the `return unless` since we don't use it if that condition is met 🤔 not a big deal at all though

PR #402, opened 2025-01-15
add user uuid to activity report

nathanstitt:
Requested by researchers so they can easily match OpenStax users to activity in Qualtrics


chrisbendel (approved):
thanks nathan!!

PR #404, opened 2026-07-16
Support a range of months in the activity report

nathanstitt:
Fixes the `too large http post` error from `rake report:activity[72]`, and adds range support so large spans can be pulled in chunks.

## The error

`get_users` plucked `user_id` from launches without `DISTINCT`. A participant appears once per launched stage, so a few thousand participants became tens of thousands of uuids, each added as its own `uuid:...` term to the Accounts search query — that oversized query is what got rejected. `UserInfo.for_uuids` chunks by 10, but chunking duplicates just yields many redundant chunks.

`UserNotifications#deliver_welcomes` already de-dupes this way; `LearnerActivityReport` was the one caller that did not.

Worth noting the range alone would not have reliably fixed this — a dense 12-month window could still overflow. The de-dupe addresses the cause; the range gives control over the span.

## Range support

```
rake report:activity[12-24]   # between 12 and 24 months ago
rake report:activity[72]      # everything since 72 months ago, as before
```

Bounds may be given in either order. The window is half-open (`>= 24.months.ago`, `< 12.months.ago`) so adjacent ranges tile without double-counting a launch on the boundary. Malformed input raises rather than silently reporting the wrong window.

## Testing

Added `spec/services/learner_activity_report_spec.rb` covering the window boundaries, order-insensitivity, back-compat, and that `for_uuids` receives one entry per participant rather than one per launch.

I was unable to run the suite locally — no gems installed for Ruby 3.3 in my environment — so CI is the first real run. I did verify the `month_bounds` parsing standalone across 14 inputs; the SQL and de-duping are unverified.

### DIFF
--- a/backend/app/services/learner_activity_report.rb
+++ b/backend/app/services/learner_activity_report.rb
@@ -3,9 +3,28 @@
 require 'csv'
 
 class LearnerActivityReport
+  DEFAULT_MONTHS_AGO = 1
+
   def initialize(months_ago:)
     @months_ago = months_ago
     @user_uuids = []
+  end
+
+  MONTHS_AGO_FORMAT = /\A\s*(\d+)\s*(?:-\s*(\d+)\s*)?\z/
+
+  # Accepts either a single bound ("72" => everything since 72 months ago) or a
+  # range ("12-24" => the window between 12 and 24 months ago).  Range bounds may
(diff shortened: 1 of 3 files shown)


nathanstitt:
Going to merge so @debshila can get data out of Kinetic

